PR #9, opened 2014-07-08
Condition runST import in Data/Primitive/Array.hs

kini:
runST is not used in this file on GHC 7.2.x and higher, hence importing
it from Control.Monad.ST produces a warning during compilation.  This
commit makes the warning go away.


cartazio:
@dolio @hvr why's this PR still open? does anyone know?

PR #10, opened 2014-07-25
fix _FILE_OFFSET_BITS redefined warning on Solaris/x86

kgardas:
The issue is that string.h header on Solaris includes somehow
/usr/include/sys/feature_tests.h which tests if _FILE_OFFSET_BITS
is defined and if not, then it defines it to 32 if we're compiling
32 bit code (x86). This is simply wrong since we'd like to have it
defined to 64. The issue is solved by including ghcconfig.h first
which defines _FILE_OFFSET_BITS to 64 and feature_tests.h is later
OK with that.

### DIFF
--- a/cbits/primitive-memops.c
+++ b/cbits/primitive-memops.c
@@ -1,3 +1,4 @@
+#include <ghcconfig.h>
 #include <string.h>
 #include "primitive-memops.h"
 


bgamari:
@hvr, has this been fixed properly yet?

dolio:
Pinging @hvr on this again.

cartazio:
@kgardas @hvr does this patch still need to be merged in?

kgardas:
I'm not GHC user on Solaris anymore, so I don't know what's the status of GHC on the latest Solarises. I personally don't need that anyway. Thanks!

chessai:
I'm going to close this. Solaris users are a rarity so I'm unsure how we would even verify its correctness. If anyone runs into this issue we can re-open it. @hvr is extremely busy and this issue isn't that important from a userbase standpoint, so no point in waiting for whatever correct fix was in mind.

PR #19, opened 2015-01-01
Split off PrimMonadInternal haskell/vector#65

snoyberg:
see haskell/vector#65

### DIFF
--- a/Control/Monad/Primitive.hs
+++ b/Control/Monad/Primitive.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE CPP, MagicHash, UnboxedTuples, TypeFamilies #-}
+{-# LANGUAGE FlexibleContexts, FlexibleInstances, UndecidableInstances #-}
 
 -- |
 -- Module      : Control.Monad.Primitive
@@ -13,6 +14,7 @@
 
 module Control.Monad.Primitive (
   PrimMonad(..), RealWorld, primitive_,
+  PrimMonadInternal(..),
   primToPrim, primToIO, primToST,
   unsafePrimToPrim, unsafePrimToIO, unsafePrimToST,
   unsafeInlinePrim, unsafeInlineIO, unsafeInlineST,
@@ -28,7 +30,27 @@ import GHC.IOBase ( IO(..) )
 #endif
 import GHC.ST     ( ST(..) )
 
--- | Class of primitive state-transformer monads
+import Control.Monad.Trans.Class (lift)
+import Data.Monoid (Monoid)
+
+import Control.Monad.Trans.Identity ( IdentityT)
+import Control.Monad.Trans.List     ( ListT    )
+import Control.Monad.Trans.Maybe    ( MaybeT   )
+import Control.Monad.Trans.Error    ( ErrorT, Error)
+import Control.Monad.Trans.Reader   ( ReaderT  )
+import Control.Monad.Trans.State    ( StateT   )
+import Control.Monad.Trans.Writer   ( WriterT  )
+import Control.Monad.Trans.RWS      ( RWST     )
+
+#if MIN_VERSION_transformers(0,4,0)
+import Control.Monad.Trans.Except   ( ExceptT  )
+#endif
+
+import qualified Control.Monad.Trans.RWS.Strict    as Strict ( RWST   )
+import qualified Control.Monad.Trans.State.Strict  as Strict ( StateT )
+import qualified Control.Monad.Trans.Writer.Strict as Strict ( WriterT )
+
+-- | Class of monads which can perform primitive state-transformer actions
 class Monad m => PrimMonad m where
   -- | State token type
   type PrimState m
@@ -36,7 +58,12 @@ class Monad m => PrimMonad m where
   -- | Execute a primitive operation
   primitive :: (State# (PrimState m) -> (# State# (PrimState m), a #)) -> m a
 
-
+-- | Class of primitive monads for state-transformer actions.
+--
+-- Unlike 'PrimMonad', this typeclass requires that the @Monad@ be fully
+-- expressed as a state transformer, therefore disallowing other monad
+-- transformers on top of the base @IO@ or @ST@.
+class PrimMonad m => PrimMonadInternal m where
   -- | Expose the internal structure of the monad
   internal :: m a -> State# (PrimState m) -> (# State# (PrimState m), a #)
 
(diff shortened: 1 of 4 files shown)


cartazio:
looks like the .travis formula needs to be updated to  the modern @hvr style https://github.com/hvr/multi-ghc-travis/blob/master/.travis.yml

could you swap it in?

snoyberg:
Swapped in new .travis.yml file, and rebased against newest commits as well.

snoyberg:
OK, after some tweaking, I applied a much smaller patch and got the build to pass on Travis. I was a bit surprised to notice that the test suites are *not* being run by Travis, is this intentional?

cartazio:
no its not, add em! test all the things! :)

snoyberg:
Ahh, I see why the tests aren't included: I'm the only person who's written any tests, and I neglected to add testing to Travis. OK, let's see if this latest changes works...

snoyberg:
OK, tests are running and passing. Now that all of *that* is out of the way, I think we can discuss the actual change itself here. I'm up for any bikeshedding on the name of the new typeclass, but other than that I think our hands are tied on what can be changed, since we should leave all method names and signatures the same to avoid unnecessary breakage.

cartazio:
hrm, how about we add something like `liftST :: PrimMonad m => ST (PrimState m) a -> m a`  to `Control.Monad.Primitive`? 

I dont have a *strong* objection to the *Internal name, though it is kinda ugly for an exported name.

I guess I'm mildly bothered by the fact that its still called PrimMonad, even though now its just for historical/compatibility reasons, but we've already established that this style design is needed if we want to minimize breakage.

As written, i guess it looks fine, though it'd be good to have some feedback from a few more folks:
paging @hvr @ekmett  @dolio  and @erikd  et al

cartazio:
oh, i'm noticing that this doesnt have the full multi-ghc travis formula btw

snoyberg:
I *think* `liftST` is just `primToPrim`, though I have no problem exposing `liftST` as another synonym. I agree that `MonadST` would be a better name, but am strongly opposed to changing the name at this point. I suppose one thing we could do is hide the entire implementation of `PrimMonad`, and export a ConstraintKind, e.g.:

    type MonadST m = PrimMonad m

which may achieve much the same goal. (FWIW, I'm *not* a fan of the `World` associated type name; I prefer `PrimState`.)

cartazio:
i dont think we need to do
```
type MonadST m = PrimMonad m

```

we would be better served (at least while pre 7.10 support matters)
by 
```class PrimMonad m => MonadST m where
instance PrimMonad m => MonadST m where
```
or something like that, I think?

snoyberg:
Yeah, that seems fine to me.

cartazio:
(though i do agree the UX of the type synonym approach IS better, so maybe that piece should be a bit more forward looking anyways?)

dolio:
I don't think `MonadST` is a better name than `PrimMonad`, since things built on `IO` are included as well. That's technically true of the existing `MonadST`, but I actually think that makes `MonadST` a worse name, strictly speaking.

I'm very opposed to including it for (fairly) seamless replacement of monad-st if Ed thinks that's sensible, though.

cartazio:
hrmm, you may have a good point though.

cartazio:
I guess i'm just thinking PrimMonad is no longer a *good/accurate* name and I'm fishing for better ideas :)

ekmett:
I'd rather not just start randomly recoloring the bikeshed and breaking every user for no real non-cosmetic reason.

cartazio:
fair enough, i was just trying to think outloud. paint coloring ideas withdrawn.

snoyberg:
I think the only thing really open to bikeshedding then is the name of the new typeclass. Are there any objections to the current name (PrimMonadInternal) or suggestions for better ones? I don't feel strongly about the name at all, and am happy to have it changed.

Otherwise, is there anything blocking this getting merged and released?

ekmett:
I have some concerns about the interaction between nondeterminism and modules like http://hackage.haskell.org/package/vector-0.10.12.2/docs/Data-Vector-Generic-Mutable.html that I'd very much like to work through first.

ekmett:
For example a MStream m a could never violate the linearity of the result vector before, but with this change it can, so far more of the API may wind up behind PrimMonadInternal than you might think.

ekmett:
Re constraint kinded aliases, they currently require users to turn on extensions, which makes them less useful than the old school UndecidableInstance class + instance pair, but I don't think we'd even want to do that.

ekmett:
We also need to add instances for the various types in transformers. Due to transformers having a hard Haskell 98 requirement they simply can't flow the other way.

ekmett:
Also, even PriimMonadInternal is a bit stronger than most ops require, most things only need linear or affine computation, not isomorphism to ST s.

snoyberg:
> For example a MStream m a could never violate the linearity of the result vector before, but with this change it can, so far more of the API may wind up behind PrimMonadInternal than you might think.

Can you give an example of a problematic case? I think that might help solidify the discussion.

> We also need to add instances for the various types in transformers.

Can you clarify what you mean by this? Which instances need to be added?

ekmett:
> Can you give an example of a problematic case?

That was what I wanted to take the time to either reassure myself didn't exist, or find. ;)

That said, let me try to construct one off the cuff:

Consider a world where we have

```haskell
instance PrimMonad m => PrimMonad (ListT m)
```

Now, lets look at 

```haskell
transform :: (PrimMonad m, MVector v a) => (MStream m a -> MStream m a) -> v (PrimState m) a -> m (v (PrimState m) a)
transform f v = fill v (f (mstream v))
````

This will refill the vector after modifying it with the MStream calculation.

Previously, `MStream` was effectively working in a world where 'm' must be isomorphic to `ST s` for some `s` by dint of the `PrimMonad` constraint, but now a larger class of effects is possible here! 

Before, the fact that the 'future' of the calculation was linear, we could only ever 'fill' once, and each step we took filling it in was only going to be visited (at most) once, we could throw an exception in the modified stream, but that is it. 

But now in particular with `ListT IO` effects we could choose to, say, give back multiple answers per entry in the vector, but we only have the one mutable vector to write all these changes into.

Things like that are I'm not in a hurry to rush in and say "oh this more generalized form of PrimMonad is definitely okay" because the entire API of `vector` was predicated on the more restricted notion of `PrimMonad`, and we're changing that underlying assumption, so I'm expecting there to be big gaping huge problems until I've had a chance to reassure myself to the contrary.

It seems obvious to me that monads that only use the future of the calculation in a linear (or affine) fashion are okay. It is the ones where we introduce a notion of continuation or non-determinism where it seems to me that we'll run into problems if we aren't careful.

>> We also need to add instances for the various types in transformers.

> Can you clarify what you mean by this? Which instances need to be added?

It strikes me that the very point of allowing a more general type of `PrimMonad` would be to encompass a larger suite of effects, e.g. state, environments, writer logs, etc. So once you open up the world of `PrimMonad` beyond `IO` and `ST s`, someone needs to supply the instances for the most common source of such extensions. Either `primitive` has to depend on `transformers` or `transformers` has to depend on `primitive`, but `transformers` lives under a doctrine that keeps its code Haskell 98 / 2010, so the dependency between them can't viably be flipped the other way. 

Say what you will about `primitive`, but Haskell 98 / 2010 it ain't.

**Comment by dolio** [comment]
I can already write `transform` with exactly the same signature, except with a `MonadST m` constraint (and `World`), today (barring `vector` not exporting enough stuff for me to do it). It's just not pre-written.

I can agree that it might be confusing if you weren't thinking about what such an operation would do when nondeterministically returning multiple results for each element, and that it isn't likely to be useful to do, but I don't really think that's a compelling reason to make `PrimMonad` an incidental barrier to doing it.

**Comment by ekmett** [comment]
I offered up that example because it was the first one I thought of.

The main concern for me is that there _is_ a "correct" implementation of many of these operations that works for an arbitrary `PrimMonad` even in the presence of multiple uses of the success continuation; it is just dreadfully inefficient. 

The loss of canonicity for the combinators offered is the main thing that bugs me here.

I'm mostly just trying to work through the ramifications of the change, and understand what it is we're giving up to get the new power we're offering users.

In the end if we collectively decide to move forward, then so be it, but I'd rather not rush ahead and be blindsided with an issue like that without at least considering it.

**Comment by ekmett** [comment]
The instance I have in monad-st isn't something I'm proud of, it is a hack I keep waiting to be called on.

**Comment by snoyberg** [comment]
Is there ever the possibility of a competing instance? Is there ever an example of an invalid instance?

**Comment by ekmett** [comment]
Yes.

Consider any case where you have a monad that happens to have a parameter of kind * -> * before the final argument, but where that parameter isn't a 'base' monad it is transforming.

We tend not to have many of these by convention, but they can and do exist.

**Comment by cartazio** [comment]
What if instead of a catch all, we provide something like 

```
data PrimT (wld::*) (m ::*-> *) (a :: *) where
   PrimT :: ( State# wld -> (# State# wld, m a #) )-> PrimT wld m a 
```
 and provide a `MonadTrans` instance for `PrimT` ?

would that class of PrimMonad instances useful though? I think it side steps the broken linearity problem, but it seems like the least useful possible PrimMonad instance imagineable

**Comment by snoyberg** [comment]
@ekmett So in this case, what we would need is:

* A monad which performs state transforms,
* and takes a parameter of kind `* -> *`,
* but doesn't work as a transformer on top of `IO` or `ST`.

It sounds to me like someone would have to go pretty far out of their way to do this. Sure, it's possible, but highly unlikely to ever occur.

On the flip side, if we don't use the all-inclusive instance, we'll now have to deal with ugly issues of mutual dependencies (should either depend on primitive? or primitive on either?) and a bunch of CPP conditionals.

So even if it isn't the purest approach, I think my vote goes for being expedient and doing things the way they are in this commit and currently in monad-st.

**Comment by ekmett** [comment]
@snoyberg, 

I have counter-examples in real code, mostly where I have something free monad like that happens to incorporate `ST s` or `IO`, but where I'm working generically in the base functor.

It isn't _that_ far fetched.

**Comment by snoyberg** [comment]
If you could show it to me privately, I'd be really curious, I'm having trouble visualizing it from that description.

I think the question still stands though: is it worth going through all of the pain that will ensue from going for explicit instances for what still seems like a relatively uncommon case?

**Comment by ekmett** [comment]
The instance makes a huge blanket assumption that 'everything of kind (* ->
*) -> * -> *' that when applied to one argument is a Monad is a monad
transformer.

Consider something like a cps'd free monad with interleaved IO effects.

newtype F f a = F { runF :: forall r. (a -> IO r) -> (f r -> IO r) -> IO r }
It isn't a transformer over 'f'.

It can be an instance of MonadIO and could have been an instance of
PrimMonad or MonadST, but for that instance.

I'm very very hesitant to encode convention into law.

This becomes a real problem now that we have polykinds running around in
code.

newtype Region s a = Region { runRegion :: IO a }

If 's' happens to be instantiated at kind (* -> *), because it can be
instantiated at any kind, now it goes looking for a MonadTrans instance --
wat? This means that necessarily any instance a user writes will overlap
with this catch-all.

I'm willing to incur the extra dependency in any package I have that
provides monads to avoid ruling out a class of programs of unpredictable
size in a way we can't easily clean up later.

**Comment by cartazio** [comment]
OUCH. ok, thats a definite problem! 

yeah, in that case, i'd definitely be in favor of writing out all the instances by hand rather than doing the catch all the blocks interesting codes!

**Comment by JohnLato** [comment]
Slightly out-of-scope on this PR, but this is a really common problem that
reappears in many guises.  Writing out instances by hand now very often
saves a lot of painful work later.

**Comment by snoyberg** [comment]
OK, I know the losing side of an argument when I see it ;). I'll update this PR... and then get ready for a whole bunch of future PRs to get PrimMonad instances added.

**Comment by snoyberg** [comment]
So along these lines: @ekmett how do you feel about adding the PrimMonad instance for ExceptT in transformers-compat, instead of adding a dependency from primitive on transformers-compat?

**Comment by snoyberg** [comment]
OK, it's pushed and Travis is passing. (I rebased slightly vs the previous set of commits to put the change under discussion as the last commit, and the other housekeeping stuff around Travis lower in the stack.) I think Edward still wanted to think about linearity a bit, but so far I'm not hearing anyone complain about the name PrimMonadInternal.

**Comment by ekmett** [comment]
I'm okay with tweaking transformers-compat to do the right dance to
minimize the dependencies from primitive.

File: Control/Monad/Primitive.hs [hunk 1]
```diff
@@ -51,51 +78,106 @@ primitive_ f = primitive (\s# ->
 instance PrimMonad IO where
   type PrimState IO = RealWorld
   primitive = IO
-  internal (IO p) = p
   {-# INLINE primitive #-}
+instance PrimMonadInternal IO where
+  internal (IO p) = p
```
cartazio: should this maybe be written as 
`internal = \(IO p)-> p` 
so that it has the same saturation/inlining behavior as other instances?

cartazio: woops, nevermind, the ST instance is in the same style

Shimuuar:
I want to bikeshed a little. `PrimMonadInternal` is not very good name. This type class means that `m` is isomorphic to `ST` for some `s`. Maybe something like `StLikeMonad` would be better?

Another issue I want to bring up there's no way to constrain monad at the base of stack. Most importantly to specialize it to IO. It becomes problem if most of the library could use any PrimMonad but some need to do IO. Concrete example is mwc-random. PRNG initialization requires IO and rest of library doesn't. Only solution I can think of is to add `(MonadIO m, MonadPrim m)` constaraint but it feels somewhat redundant.

dolio:
For the latter, you can use `PrimState m ~ RealWorld`. But that's longer than `MonadIO`.

I suppose we could include an alias for `(PrimMonad m, PrimState m ~ RealWorld)` if that is actually expected to be widely useful. I'm not sure it will be, though.

cartazio:
i'm inclined to agree with @dolio ,

ekmett:
It also has the disadvantage of not being able to be partially applied,
unless we do the usual undecidable/flexible class/instance pair.

ekmett:
Er.. by partially applied I mean, 'left unapplied' as a type of kind (* ->
*) -> Constraint.

cartazio:
@ekmett  clarified what he meant on irc
```
[15/01/20-23:56:10]  <carter>	edwardk: i cant quite parse your gh remark on the vector thread
[15/01/20-23:56:31]  <edwardk>	don't remember my remark
[15/01/20-23:56:42]  <edwardk>	oh
[15/01/20-23:57:00]  <edwardk>	re: type MonadIO m = (PrimMonad m, PrimState m ~ RealWorld)
[15/01/20-23:57:01]  <edwardk>	?
[15/01/20-23:57:29]  <edwardk>	class (PrimMonad m, PrimState m ~ RealWorld) => MonadIO m; instance (PrimMonad m, PrimState m ~ RealWorld) => MonadIO m
[15/01/20-23:57:38]  <edwardk>	the latter is a more robust way of saying the former
[15/01/20-23:57:50]  <edwardk>	the former requires an extension on the behalf of the user before 7.10
[15/01/20-23:58:08]  <edwardk>	but it _also_ has the problem that you can't talk about 'MonadIO :: (* -> *) -> Constraint' as a type in its own right
[15/01/20-23:58:22]  <edwardk>	its only a 'thing' when its applied to something MonadIO IO :: Constraint
[15/01/20-23:59:06]  <carter>	ohh
[15/01/20-23:59:12]  <edwardk>	in the same sense as type Flip f a b = f b a   -- is only a 'thing' when you have given it all 3 arguments. you can't write  instance Monad (Flip Either x)
```

which i agree with

Shimuuar:
Yes. I agree that PrimState m ~ RealWorld is not best idea. It's also quite
cryptic. So probably (MonadPrim m, MonadIO m) is best solution. MonadIO
type synonym will also clash with MonadIO from trasformers

Any opinion on naming of PrimMonadInternal?

cartazio:
i dont really care what we name it, but `PrimMonadBase` or `PrimMonadSimple` are probably viable too right?

Shimuuar:
I don't particulary care either. I just think that PrimMonadInternal is
actively misleading. Your variants are fine too. Even better probably since
they mention PrimMonad in their name

cartazio:
we should probably also expose a pretty version of `primitive`  that is prettier as a derived combinator.

I think @dolio  will do that a few other bits of cleanup and ship this soon (we talked through this and a few related action items at compose conf)

snoyberg:
It looks to me like this PR has officially stalled. Anything we should do to un-stall it?

dolio:
I'll talk to Ed tomorrow about any remaining objection he has.

Other than that, there seems to be some dislike of the name. But I don't have a better one, and neither does anyone else, I think. Does anyone like the other two names Carter suggested better than the current one? Or have something better? Otherwise I'll probably just merge it.

Shimuuar:
I like Carter's `MonadPrimBase` much better than `MonadPrimInternal`. Internal usually used to describe private parts of API but this is public API. So I think `PrimMonadInternal` is misleading name

dolio:
How does `PrimBase` strike everyone? I think it's nicely short, and it isn't a travesty that it doesn't have the word "monad" in it.

Don't bother revising the pull request. I'll just merge it and then we can rename on top of it.

Shimuuar:
It's nice too

cartazio:
that works for me.

snoyberg:
No objection from me.

dolio:
@cartazio By the way, I just realized that `primToPrim` is the nicer form of `primitive` that you wanted. It can lift any `PrimBase` into a stack based on that base.

It might be difficult to notice that. Perhaps we should mention that in the comment/documentation?

cartazio:
@dolio, YES, good point, we should probably document it! Would it be worth having some sort of `liftPrim` alias?

cartazio:
but yeah, one way or another, we should probably make `primToPrim` a bit more prominent in the docs

PR #21, opened 2015-01-07
Add evalPrim

treeowl:
Generalize `Control.Exception.evaluate`.


cartazio:
looks like you need to add a teeny bit more cpp https://travis-ci.org/haskell/primitive/jobs/46224608

PR #27, opened 2015-07-02
Use size_t and ptrdiff_t instead of int in cbits

dolio:
The current implementation overflows if you try to memset a large enough array (2^32 or greater).


cartazio:
lgtm (though i've not tested it myself :) )

PR #33, opened 2015-08-13
Fix a GHC 7.10 warning.

23Skidoo:


glguy:
In the past Edward has mentioned to me that old versions of the MIN_VERSION_* macros could not be negated without being surrounded by parentheses.

    #if !(MIN_VERSION_base(4,8,0))

23Skidoo:
@glguy 
> In the past Edward has mentioned to me that old versions of the MIN_VERSION_* macros could not be negated without being surrounded by parentheses.

`!MIN_VERSION_foo` is already used in other places in the code without parentheses:

```
$ ag MIN_VERSION 
Data/Primitive/Internal/Compat.hs
[...]
31:#if !MIN_VERSION_base(4,2,0)
35:#if !MIN_VERSION_base(4,7,0)
```

Looking at Cabal code, I see that that the missing parentheses bug was fixed in 2009: https://github.com/haskell/cabal/commit/ff9f0aaa44a73f281846fa475570878c56c656a3.

I can of course update this PR if that what's required to get it merged.

glguy:
Given that the package requires Cabal >= 1.10 and this wasn't an issue in 1.10 I retract my comment.

Issue #20, opened 2015-01-03
Support for new SmallArray# family of types and functions in GHC 7.10


jberryman:
I suppose the options are:

  - break compatibility with old `base` and support `SmallArray#` as a wrapper in the usual way

  - export an opaque `SmallArray` type that conditionally wraps `SmallArray#` else `Array#` for old versions

  - Make `Array`/`MutableArray` opaque (causing breakage for some, but probably not most users) something like:

        data Array a = Array (Array# a) 
        #if __GLASGOW_HASKELL__ >= 710
                     | SmallArray (SmallArray# a)
        #endif

    ...and then just provide an additional `newSmallArray` function, plus support small arrays in existing `copy/clone/etc` functions.

Other ideas? I'd be happy to try to submit a patch if we can figure out the best approach.

jberryman:
Now that I'm looking at my own code, I would very much prefer the third option as a user; otherwise it's quite a pain to even try out small arrays and I think users will end up not using `primitive` and just implementing this proposal themselves.

Then I suppose exposing the constructor(s) in an `Internal` module is the best choice. Thoughts?

cartazio:
@jberryman  pring

cartazio:
i do think that anything we add to the primitive package shoulnd't be such that it hinders the performance minded users

ekmett:
It strikes me that it probably belongs as its own `SmallArray` type.

dolio:
Small arrays have been merged.

PR #36, opened 2015-10-28
Add support for small arrays

dolio:
This addresses issue #20.

How do we feel about the loops? (Note, I checked the core, and they look just as good as named definitions; so that's not an issue.)


cartazio:
which loops are you referring to? (that said, trying to work through understanding the code :) )

that does remind me:  ghc's  batch array  primops dont play nice with the scheduler large arrays, and either at the ghc level, or perhaps at the primitive layer that makes sure we yield at least once every few million entries or something?

PR #39, opened 2016-07-25
Add Prim instances for Ptr and FunPtr

### LINKED ISSUE #31: Add instance Prim (Ptr a)
Like it says on the tin.

We have a superfluous `Addr` type in here, which similarly wraps `Addr#`, but we could add an instance for `Ptr a` fairly readily.

RyanGlScott:
Fixes #31.


dolio:
Nice.

PR #40, opened 2016-07-25
Add ioToPrim, stToPrim, and unsafe counterparts

### LINKED ISSUE #26: Add ioToPrim, stToPrim
There is no nice way to lift an `ST s` action into `PrimMonad` without a big type signature on `primToPrim`. We can go the other way, `primToST`, `primToIO` (with `PrimBase`) but there is no `ioToPrim` or `stToPrim`, able to be used on any `PrimMonad`.

RyanGlScott:
Fixes #26.


cartazio:
I suppose it's too early days yet to add steToPrim ;)

PR #47, opened 2016-10-22
Add a type safe wrapper around ArrayArray#

### LINKED ISSUE #46: Add ArrayArrays?
`ArrayArray#` and `MutableArrayArray#` have been around since ghc-prim 0.3.1.0, and adding `Data.Primitive.ArrayArray` should be fairly mechanical. Is there a reason why it hasn't been done? Is it because they're even less type-safe than other primitive operations?

dolio:
Potentially resolves #46.

Topics for discussion:
- This wrapper is type safe, homogeneous arrays of things that are wrappers around unlifted types, is there real cause for a type-unsafe, heterogeneous interface?
- This GHC API only goes back to 7.8 I believe (travis should tell us for sure). What do we do about that? Break support for 7.6 and earlier? Conditionally provide the API based on GHC version? Something else?
- I put the class that identifies unliftable things in the module that defines the unlifted arrays. It could instead go in the same place that the `Prim` class is. This would mean that each of the array types would import it and define their own instances, but that means they all have to import `unsafeCoerce#`, instead of just the one module doing that.

### DIFF
--- a/Data/Primitive/UnliftedArray.hs
+++ b/Data/Primitive/UnliftedArray.hs
@@ -1,0 +1,362 @@
+{-# Language MagicHash #-}
+{-# Language UnboxedTuples #-}
+
+-- |
+-- Module      : Data.Primitive.UnliftedArray
+-- Copyright   : (c) Dan Doel 2016
+-- License     : BSD-style
+--
+-- Maintainer  : Libraries <libraries@haskell.org>
+-- Portability : non-portable
+--
+-- GHC contains three general classes of value types:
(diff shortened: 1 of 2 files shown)


dolio:
Evidently `ArrayArray#` goes all the way back to 7.4.2, so I propose that we just drop things before that. That happens to be the same cutoff as vector.

cartazio:
Sound good to me.

cartazio:
Mind you I've not looked at the associated code

PR #49, opened 2016-11-16
Version bump, document new additions in changelog.

dolio:


hvr:
btw, it's not obvious from the changelog why a major version bump was performed (which deserves to be made apparent in a changelog, IMHO :-) )

cartazio:
Probably the dropped support for 7.4?

hvr:
@cartazio why would dropping support for GHC 7.4/base-4.5  require a major version bump? :-)

dolio:
I could lower the version increment. I simply thought that adding two completely new modules plus dropping support for certain versions was a fairly significant change.

hvr:
Well, the problem here is packages depending on `primitive` which rely on PVP contract. Those will have to go through the API review process (due to a major version increment being signalled) to decide whether it's safe to allow that new version of `primitive`.

However, adding a new module is merely an API addition (i.e. minor version bump), while constraining the version range of supported dependencies is totally benign (as it can't break the semantics or compilation of any existing API consumer) and would at best require a patch-level increment.

So from the point of view of the PVP, this is not a "significant change" at all, but merely a (significant) API  addition.

cartazio:
Makes sense.

As long as it's easy to see this stuff via the change log I guess that's
all good

PR #57, opened 2017-06-01
Various fixes

bgamari:
Fix a variety of papercuts I recently encountered.

### DIFF
--- a/Data/Primitive/Addr.hs
+++ b/Data/Primitive/Addr.hs
@@ -22,12 +22,16 @@ module Data.Primitive.Addr (
   indexOffAddr, readOffAddr, writeOffAddr,
 
   -- * Block operations
-  copyAddr, moveAddr, setAddr
+  copyAddr, moveAddr, setAddr,
+
+  -- * Conversion
+  addrToInt
 ) where
 
 import Control.Monad.Primitive
 import Data.Primitive.Types
 
+import Numeric
 import GHC.Base ( Int(..) )
 import GHC.Prim
 
@@ -100,3 +104,7 @@ setAddr :: (Prim a, PrimMonad m) => Addr -> Int -> a -> m ()
 {-# INLINE setAddr #-}
 setAddr (Addr addr#) (I# n#) x = primitive_ (setOffAddr# addr# 0# n# x)
 
+-- | Convert an 'Addr' to an 'Int'.
+addrToInt :: Addr -> Int
+{-# INLINE addrToInt #-}
+addrToInt (Addr addr#) = I# (addr2Int# addr#)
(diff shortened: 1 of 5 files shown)


Data/Primitive/Addr.hs
```diff
@@ -100,3 +107,11 @@ setAddr :: (Prim a, PrimMonad m) => Addr -> Int -> a -> m ()
 {-# INLINE setAddr #-}
 setAddr (Addr addr#) (I# n#) x = primitive_ (setOffAddr# addr# 0# n# x)
 
+-- | Convert an 'Addr' to an 'Int'.
+addrToInt :: Addr -> Int
+{-# INLINE addrToInt #-}
+addrToInt (Addr addr#) = I# (addr2Int# addr#)
```
cartazio: wouldn't Word be more correct / robustly portable?
(granted both options are sad and very very very ghc RTS/semantics specific, so i can't complain too much :) )

Data/Primitive/Addr.hs
```diff
@@ -100,3 +107,11 @@ setAddr :: (Prim a, PrimMonad m) => Addr -> Int -> a -> m ()
...
+
+-- | Produce a string representation of an 'Addr'. Useful while debugging.
+showAddr :: Addr -> String
+showAddr addr = "0x" ++ showHex (addrToInt addr) ""
```
cartazio: what happens on systems where this looks like a negative number?

Data/Primitive/ByteArray.hs
```diff
@@ -70,10 +70,13 @@ newPinnedByteArray (I# n#)
...
--- | Create a /pinned/ byte array of the specified size and with the give
+-- | Create a /pinned/ byte array of the specified size and with the given
 -- alignment. The garbage collector is guaranteed not to move it.
 newAlignedPinnedByteArray
-  :: PrimMonad m => Int -> Int -> m (MutableByteArray (PrimState m))
+  :: PrimMonad m
+  => Int  -- ^ size
+  -> Int  -- ^ alignment
+  -> m (MutableByteArray (PrimState m))
 {-# INLINE newAlignedPinnedByteArray #-}
```
cartazio: good clarification and seems correct :)

Data/Primitive/Addr.hs
bgamari: Indeed it would. Fixing.

bgamari: Actually, never mind; I think given that GHC exposes `addr2Int#` and not `addr2Word#` I'd rather follow the prior art.

Data/Primitive/Addr.hs
bgamari: Fixing.

bgamari:
I'm going to close this and open another from another branch of my fork.

PR #60, opened 2017-07-06
Add PrimMonad instances for remaining monad transformers

RyanGlScott:
There were missing `PrimMonad` instances for some monad transformer types:

* `ContT`
* `AccumT` and `SelectT`, which were added to `transformers` in 0.5.3

This PR adds them.

### DIFF
--- a/Control/Monad/Primitive.hs
+++ b/Control/Monad/Primitive.hs
@@ -41,6 +41,7 @@ import Control.Monad.Trans.Class (lift)
 import Data.Monoid (Monoid)
 #endif
 
+import Control.Monad.Trans.Cont     ( ContT    )
 import Control.Monad.Trans.Identity ( IdentityT (IdentityT) )
 import Control.Monad.Trans.List     ( ListT    )
 import Control.Monad.Trans.Maybe    ( MaybeT   )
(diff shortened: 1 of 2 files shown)


cartazio:
@RyanGlScott  ContT probably needs a warning on the TIN, just like LogicT or ListT would, not familiar with the latter two

andrewthad:
I agree that `ContT`'s instance should have a warning. `AccumT` is definitely safe to use, and I have no idea what `SelectT` does.

RyanGlScott:
My apologies, I didn't realize the fires I was stoking here :)

I didn't see a warning attached to the existing `ListT` instance, so perhaps that's why I didn't notice. Can you give a template for the the warning I should put in the Haddocks?

cartazio:
I kinda think the decision when we made primonad stackable was to not
provide conTT or listT.

But maybe having a .unsafe module with those as orphans is ok?

cartazio:
The main goal is to make sure that backtracking stuff and associated bugs
are user beware ..

cartazio:
Maybe the right approach is to Improve the readme and have some remarks in
the docs

I’ll see if I can draft something this afternoon

PR #62, opened 2017-08-02
ByteArray instances

bgamari:
Add `Eq`, `Ord`, and `Show` instances for `ByteArray`.

### DIFF
--- a/Data/Primitive/ByteArray.hs
+++ b/Data/Primitive/ByteArray.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE CPP, MagicHash, UnboxedTuples, UnliftedFFITypes, DeriveDataTypeable #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 
 -- |
 -- Module      : Data.Primitive.ByteArray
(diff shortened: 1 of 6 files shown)


dolio:
I'm not sure where `I#` is coming from in the newer versions, but I guess it's not there in older ones?

By the way, you could use `Data.Primitive.sizeOf` instead of `I# (sizeOf# ...)` to not have to worry about this in some places. :)

Data/Primitive/ByteArray.hs
```diff
@@ -262,3 +281,49 @@ instance Typeable s => Data (MutableByteArray s) where
   toConstr _ = error "toConstr"
   gunfold _ _ = error "gunfold"
   dataTypeOf _ = mkNoRepType "Data.Primitive.ByteArray.MutableByteArray"
+
+instance Show ByteArray where
+  showsPrec _ ba =
+      showString "ByteArray [" . go 0
```
hvr: given the `IsList` instance, I think it would be better to have `Show` emit something that would parse on the REPL, i.e.

    [0xca,0xfe,0xbe,0xef]

rather than

    ByteArray [0xca,0xfe,0xbe,0xef]

bgamari: Fair point; didn't really think through that one.

bgamari:
How does this look?

dolio:
Looks good.

PR #63, opened 2017-08-14
Implement resizeMutableByteArray

bgamari:
This provides access to the `resizeMutableByteArray#` primop added in GHC 7.10.

Note that this operation, like the underlying primop, doesn't preserve pinned-ness. Another alternate design would be to pin the new array if the original was pinned. However, it is unfortunately not possible to provide these semantics consistently across GHC versions since `isByteArrayPinned#` was only added in GHC 8.0.

### DIFF
--- a/Data/Primitive/ByteArray.hs
+++ b/Data/Primitive/ByteArray.hs
@@ -17,6 +17,7 @@ module Data.Primitive.ByteArray (
 
   -- * Allocation
   newByteArray, newPinnedByteArray, newAlignedPinnedByteArray,
+  resizeMutableByteArray,
 
   -- * Element access
   readByteArray, writeByteArray, indexByteArray,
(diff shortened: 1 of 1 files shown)


bgamari:
Any thoughts on this?

andrewthad:
This seems like an obvious win. I'm in favor of merging this.

treeowl:
I'd personally favor doing whatever is best, whether or not it works with old GHC versions.

treeowl:
On the other hand, this is `primitive`, so we probably should just expose the underlying primops and let people do what they like with them.

andrewthad:
This PR does all three: it does what is best, it works with old GHCs, and it just wraps the underlying primop. I'm not worried at all about the fact that it doesn't pin the new bytearray, since it's documented. Buyer beware.

treeowl:
Yes, indeed. Thanks, @bgamari.

Data/Primitive/ByteArray.hs
```diff
@@ -100,6 +102,45 @@ sameMutableByteArray :: MutableByteArray s -> MutableByteArray s -> Bool
...
+                 -> m (MutableByteArray (PrimState m))
+{-# INLINE resizeMutableByteArray #-}
+#if __GLASGOW_HASKELL__ >= 710
+resizeMutableByteArray (MutableByteArray arr#) (I# n#)
+  = primitive (\s# -> case resizeMutableByteArray# arr# n# s# of
+                        (# s'#, arr'# #) -> (# s'#, MutableByteArray arr'# #))
+#else
+resizeMutableByteArray arr n
+  = do arr' <- newByteArray n
+       copyMutableByteArray arr 0 arr' 0 (min (sizeofMutableByteArray arr) n)
```
hvr: for the record, this was identified to be incorrect; c.f. #116

PR #64, opened 2017-08-15
add PrimArray, Arr class

winterland1989:
This is the `PrimArray` patch as @bgamari requested, This patch also offer polymorphric `Arr` class which unify boxed and unboxed array interface. Please review carefully because there may be bugs ; ) 

~~I haven't consider much compatibility issue when i'm doing this, so please add whatever `CPP` needed to make it work on older GHCs.~~ I have added `CPP`s and CI is happy now.

~~I'm planning to make another patch which bring `foldr/build` fusion to these arrays, please don't cut a new release before that is done.~~ I believe fusion belongs to high level vector package now, so this is the full patch already, please review and merge(if there's no objections) ASAP.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,6 +51,7 @@ install:
           $HOME/.cabal/packages/hackage.haskell.org/00-index.tar;
    fi
  - travis_retry cabal update -v
+ - cabal install cpphs
  - sed -i 's/^jobs:/-- jobs:/' ${HOME}/.cabal/config
  - cabal new-build ${TEST} ${BENCH} --dep
 
--- a/Data/Primitive/Array/Checked.hs
+++ b/Data/Primitive/Array/Checked.hs
@@ -70,7 +70,7 @@ import GHC.Ptr (Ptr(..))
 
 check :: String -> Bool -> a -> a
 check _      True  x = x
-check errMsg False _ = throw (IndexOutOfBounds $ "Data.Array.Checked." ++ errMsg)
+check errMsg False _ = throw (IndexOutOfBounds $ "Data.Primitive.Array.Checked." ++ errMsg)
 {-# INLINE check #-}
 
 newArr :: (A.Arr marr arr a, PrimMonad m, PrimState m ~ s) => Int -> m (marr s a)
--- a/Data/Primitive/Array/Checked.hs
+++ b/Data/Primitive/Array/Checked.hs
@@ -1,0 +1,225 @@
+{-# LANGUAGE TypeFamilies #-}
+
+{-|
+Module      : Data.Array.Checked
+Description : Bounded checked boxed and unboxed arrays
+Copyright   : (c) Winter, 2017
+License     : BSD3
+Maintainer  : libraries@haskell.org, drkoster@qq.com
+
+Stability   : experimental
+Portability : non-portable
+
+This module provides exactly the same API with "Data.Primitive.Array.Class", but will throw an 'IndexOutOfBounds'
+'ArrayException' on bound check failure.
+
+-}
+module Data.Primitive.Array.Checked
+  ( -- * Arr typeclass re-export
+    A.Arr
+  , RealWorld
+    -- * Bound checked array operations
+  , newArr
+  , newArrWith
+  , readArr
+  , writeArr
+  , setArr
+  , indexArr
+  , indexArrM
+  , freezeArr
+  , thawArr
+  , copyArr
+  , copyMutableArr
+  , moveArr
+  , cloneArr
+  , cloneMutableArr
+  , resizeMutableArr
+  , shrinkMutableArr
+  -- * No bound checked operations
+  , A.unsafeFreezeArr
+  , A.unsafeThawArr
+  , A.sameMutableArr
+  , A.sizeofArr
+  , A.sizeofMutableArr
+  , A.sameArr
+  -- * Boxed array type
+  , A.Array(..)
+  , A.MutableArray(..)
+  , A.SmallArray(..)
+  , A.SmallMutableArray(..)
+  , A.uninitialized
+  -- * Primitive array type
+  , A.PrimArray(..)
+  , A.MutablePrimArray(..)
+  , A.newPinnedPrimArray, A.newAlignedPinnedPrimArray
+  -- * Bound checked primitive array operations
+  , copyPrimArrayToPtr, copyMutablePrimArrayToPtr, copyMutablePrimArrayFromPtr
+  -- * Unlifted array type
+  , A.UnliftedArray(..)
+  , A.MutableUnliftedArray(..)
+  , A.PrimUnlifted(..)
+  -- * The 'ArrayException' type
+  , ArrayException(..)
+  ) where
+
+import qualified Data.Primitive.Array.Class as A
+import Control.Exception (throw, ArrayException(..))
+import Control.Monad.Primitive
+import Data.Primitive.Types
+import GHC.Ptr (Ptr(..))
+
+check :: String -> Bool -> a -> a
+check _      True  x = x
+check errMsg False _ = throw (IndexOutOfBounds $ "Data.Array.Checked." ++ errMsg)
+{-# INLINE check #-}
+
+newArr :: (A.Arr marr arr a, PrimMonad m, PrimState m ~ s) => Int -> m (marr s a)
+newArr n = check "newArr: negative size" (n>=0) (A.newArr n)
+{-# INLINE newArr #-}
+
+newArrWith :: (A.Arr marr arr a, PrimMonad m, PrimState m ~ s) => Int -> a -> m (marr s a)
+newArrWith n x = check "newArrWith: negative size" (n>=0) (A.newArrWith n x)
+{-# INLINE newArrWith #-}
+
+readArr :: (A.Arr marr arr a, PrimMonad m, PrimState m ~ s) => marr s a -> Int -> m a
+readArr marr i = do
+    siz <- A.sizeofMutableArr marr
+    check "readArr: index of out bounds"
+        (i>=0 && i<siz)
+        (A.readArr marr i)
+{-# INLINE readArr #-}
+
+writeArr :: (A.Arr marr arr a, PrimMonad m, PrimState m ~ s) => marr s a -> Int -> a -> m ()
+writeArr marr i x = do
+    siz <- A.sizeofMutableArr marr
+    check "writeArr: index of out bounds"
+        (i>=0 && i<siz)
+        (A.writeArr marr i x)
+{-# INLINE writeArr #-}
+
+setArr :: (A.Arr marr arr a, PrimMonad m, PrimState m ~ s) => marr s a -> Int -> Int -> a -> m ()
+setArr marr s l x = do
+    siz <- A.sizeofMutableArr marr
+    check "setArr: index range of out bounds"
+        (s>=0 && l>=0 && (s+l)<=siz)
+        (A.setArr marr s l x)
+{-# INLINE setArr #-}
+
+indexArr :: (A.Arr marr arr a) => arr a -> Int -> a
+indexArr arr i = check "indexArr: index of out bounds"
+    (i>=0 && i<A.sizeofArr arr)
+    (A.indexArr arr i)
+{-# INLINE indexArr #-}
+
+indexArrM :: (A.Arr marr arr a, PrimMonad m, PrimState m ~ s) => arr a -> Int -> m a
+indexArrM arr i = check "indexArrM: index of out bounds"
+    (i>=0 && i<A.sizeofArr arr)
+    (A.indexArrM arr i)
+{-# INLINE indexArrM #-}
+
+freezeArr :: (A.Arr marr arr a, PrimMonad m, PrimState m ~ s) => marr s a -> Int -> Int -> m (arr a)
+freezeArr marr s l = do
+    siz <- A.sizeofMutableArr marr
+    check "freezeArr: index range of out bounds"
+        (s>=0 && l>=0 && (s+l)<=siz)
+        (A.freezeArr marr s l)
+{-# INLINE freezeArr #-}
+
+thawArr :: (A.Arr marr arr a, PrimMonad m, PrimState m ~ s) => arr a -> Int -> Int -> m (marr s a)
+thawArr arr s l = check "thawArr: index range of out bounds"
+    (s>=0 && l>=0 && (s+l)<=A.sizeofArr arr)
+    (A.thawArr arr s l)
+{-# INLINE thawArr #-}
+
+copyArr :: (A.Arr marr arr a, PrimMonad m, PrimState m ~ s) => marr s a -> Int -> arr a -> Int -> Int -> m ()
+copyArr marr s1 arr s2 l = do
+    siz <- A.sizeofMutableArr marr
+    check "copyArr: index range of out bounds"
+        (s1>=0 && s2>=0 && l>=0 && (s2+l)<=A.sizeofArr arr && (s1+l)<=siz)
+        (A.copyArr marr s1 arr s2 l)
+{-# INLINE copyArr #-}
+
+copyMutableArr :: (A.Arr marr arr a, PrimMonad m, PrimState m ~ s) => marr s a -> Int -> marr s a -> Int -> Int -> m ()
+copyMutableArr marr1 s1 marr2 s2 l = do
+    siz1 <- A.sizeofMutableArr marr1
+    siz2 <- A.sizeofMutableArr marr2
+    check "copyMutableArr: index range of out bounds"
+        (s1>=0 && s2>=0 && l>=0 && (s2+l)<=siz2 && (s1+l)<=siz1)
+        (A.copyMutableArr marr1 s1 marr2 s2 l)
+{-# INLINE copyMutableArr #-}
+
+moveArr :: (A.Arr marr arr a, PrimMonad m, PrimState m ~ s) => marr s a -> Int -> marr s a -> Int -> Int -> m ()
+moveArr marr1 s1 marr2 s2 l = do
+    siz1 <- A.sizeofMutableArr marr1
+    siz2 <- A.sizeofMutableArr marr2
+    check "moveArr: index range of out bounds"
+        (s1>=0 && s2>=0 && l>=0 && (s2+l)<=siz2 && (s1+l)<=siz1)
+        (A.copyMutableArr marr1 s1 marr2 s2 l)
+{-# INLINE moveArr #-}
+
+cloneArr :: (A.Arr marr arr a) => arr a -> Int -> Int -> arr a
+cloneArr arr s l = check "cloneArr: index range of out bounds"
+    (s>=0 && l>=0 && (s+l)<=A.sizeofArr arr)
+    (A.cloneArr arr s l)
+{-# INLINE cloneArr #-}
+
+cloneMutableArr :: (A.Arr marr arr a, PrimMonad m, PrimState m ~ s) => marr s a -> Int -> Int -> m (marr s a)
+cloneMutableArr marr s l = do
+    siz <- A.sizeofMutableArr marr
+    check "cloneMutableArr: index range of out bounds"
+        (s>=0 && l>=0 && (s+l)<=siz)
+        (A.cloneMutableArr marr s l)
+{-# INLINE cloneMutableArr #-}
+
+resizeMutableArr :: (A.Arr marr arr a, PrimMonad m, PrimState m ~ s) => marr s a -> Int -> m (marr s a)
+resizeMutableArr marr n = check "resizeMutableArr: negative index"
+    (n>=0)
+    (A.resizeMutableArr marr n)
+{-# INLINE resizeMutableArr #-}
+
+-- | New size should be >= 0, and <= original size.
+--
+shrinkMutableArr :: (A.Arr marr arr a, PrimMonad m, PrimState m ~ s) => marr s a -> Int -> m ()
+shrinkMutableArr marr n = do
+    siz <- A.sizeofMutableArr marr
+    check "shrinkMutableArr: size out of bounds"
+        (n>=0 && n<=siz)
+        (A.shrinkMutableArr marr n)
+{-# INLINE shrinkMutableArr #-}
+
+copyPrimArrayToPtr :: (PrimMonad m, Prim a)
+                   => Ptr a
+                   -> A.PrimArray a
+                   -> Int
+                   -> Int
+                   -> m ()
+{-# INLINE copyPrimArrayToPtr #-}
+copyPrimArrayToPtr ptr arr s l = check "copyPrimArrayToPtr: index range out of bounds"
+    (s>=0 && l>=0 && (s+l)<=A.sizeofArr arr)
+    (A.copyPrimArrayToPtr ptr arr s l)
+
+copyMutablePrimArrayToPtr :: (PrimMonad m, Prim a)
+                          => Ptr a
+                          -> A.MutablePrimArray (PrimState m) a
+                          -> Int
+                          -> Int
+                          -> m ()
+{-# INLINE copyMutablePrimArrayToPtr #-}
+copyMutablePrimArrayToPtr ptr marr s l = do
+    siz <- A.sizeofMutableArr marr
+    check "copyMutablePrimArrayToPtr: index range out of bounds"
+        (s>=0 && l>=0 && (s+l)<=siz)
+        (A.copyMutablePrimArrayToPtr ptr marr s l)
+
+copyMutablePrimArrayFromPtr :: (PrimMonad m, Prim a)
+                            => A.MutablePrimArray (PrimState m) a
+                            -> Int
+                            -> Ptr a
+                            -> Int
+                            -> m ()
+{-# INLINE copyMutablePrimArrayFromPtr #-}
+copyMutablePrimArrayFromPtr marr s ptr l = do
+    siz <- A.sizeofMutableArr marr
+    check "copyMutablePrimArrayFromPtr: index range out of bounds"
+        (s>=0 && l>=0 && (s+l)<=siz)
+        (A.copyMutablePrimArrayFromPtr marr s ptr l)
--- a/Data/Primitive/Array/Class.hs
+++ b/Data/Primitive/Array/Class.hs
@@ -319,11 +319,13 @@ instance Arr SmallMutableArray SmallArray a where
     shrinkMutableArr _ _ = return ()
     {-# INLINE shrinkMutableArr #-}
 
-    sameMutableArr (SmallMutableArray smarr1#) (SmallMutableArray smarr2#) = isTrue#
 #if HAVE_SMALL_ARRAY
+    sameMutableArr (SmallMutableArray smarr1#) (SmallMutableArray smarr2#) = isTrue#
         (sameSmallMutableArray# smarr1# smarr2#)
 #else
-        (sameMutableArray# smarr1# smarr2#)
+    sameMutableArr (SmallMutableArray (MutableArray smarr1#))
+                   (SmallMutableArray (MutableArray smarr2#)) = isTrue#
+                        (sameMutableArray# smarr1# smarr2#)
 #endif
     {-# INLINE sameMutableArr #-}
 
@@ -332,10 +334,11 @@ instance Arr SmallMutableArray SmallArray a where
     sizeofMutableArr = return . sizeofSmallMutableArray
     {-# INLINE sizeofMutableArr #-}
 
-    sameArr (SmallArray arr1#) (SmallArray arr2#) = isTrue#
 #if HAVE_SMALL_ARRAY
+    sameArr (SmallArray arr1#) (SmallArray arr2#) = isTrue#
         (sameSmallMutableArray# (unsafeCoerce# arr1#) (unsafeCoerce# arr2#))
 #else
+    sameArr (SmallArray (Array arr1#)) (SmallArray (Array arr2#)) = isTrue#
         (sameMutableArray# (unsafeCoerce# arr1#) (unsafeCoerce# arr2#))
 #endif
     {-# INLINE sameArr #-}
--- a/Data/Primitive/Array/Class.hs
+++ b/Data/Primitive/Array/Class.hs
@@ -427,8 +427,8 @@ instance PrimUnlifted a => Arr MutableUnliftedArray UnliftedArray a where
     unsafeFreezeArr = unsafeFreezeUnliftedArray
     {-# INLINE unsafeFreezeArr #-}
     unsafeThawArr (UnliftedArray arr#) = primitive ( \ s0# ->
-            let (# s1#, marr# #) = unsafeThawArray# (unsafeCoerce# arr#) s0#  -- ArrayArray# and Array# use the same representation
-            in (# s1#, MutableUnliftedArray (unsafeCoerce# marr#) #)          -- so this works
+            let !(# s1#, marr# #) = unsafeThawArray# (unsafeCoerce# arr#) s0#   -- ArrayArray# and Array# use the same representation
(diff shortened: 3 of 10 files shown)


winterland1989:
After reading https://github.com/haskell/primitive/pull/63 , i think it's probable a good idea to keep pinning status in reszied array, ideas?

andrewthad:
Small life tip: if you change your `~/.gitignore` to have this section:

    [user]
        email = winterland1989@gmail.com
        name = John Doe

Then on github, your commits will be attributed to your user instead of being attributed to "handong", who is unknown to github.

andrewthad:
I've put up a PR against your branch fixing minor documentation typos: https://github.com/winterland1989/primitive/pull/1

winterland1989:
Thanks!

andrewthad:
I dislike the current `Eq` and `Ord` definitions for `PrimArray`. I would prefer to see `Eq` actually use the `Eq` constraint on the element type, as I do [here](https://github.com/andrewthad/prim-array/blob/1abf88b9eff5262a0ffc7eb609e7b23f9bd7285d/src/Data/Primitive/PrimArray.hs#L54-L58). This way, if someone wanted to use `PrimArray` of numbers whose equality uses modular arithmetic, it would still work (or really any notion of equality that allows structurally different things to be equal).

Also, for very different reasons, I do not like the overlapping instance for `Ord`. Overlapping instances  mess up instance resolution sometimes. I'll try to come up with an example of this.

I would prefer that fast versions of equality and of comparison be provided as functions but that they not be used to attempt to optimize the `Eq` and `Ord` typeclass instances. Providing `eqPrimArrayBytewise` and `comparePrimArrayBytewise` would be my preference.

Data/Primitive/Array/Checked.hs
```diff
@@ -0,0 +1,225 @@
...
+import qualified Data.Primitive.Array.Class as A
+import Control.Exception (throw, ArrayException(..))
+import Control.Monad.Primitive
+import Data.Primitive.Types
+import GHC.Ptr (Ptr(..))
+
+check :: String -> Bool -> a -> a
+check _      True  x = x
+check errMsg False _ = throw (IndexOutOfBounds $ "Data.Array.Checked." ++ errMsg)
+{-# INLINE check #-}
```
cartazio: for this module, does INLINE pragmas make sense? theres definitely no fusion going on , and to quote @thoughtpolice on the topic, if ghc isn't marking them inlinable when they're small enough, thats a ghc bug ..

Data/Primitive/Array/Class.hs
```diff
@@ -0,0 +1,485 @@
...
+-- | Bottom value (@throw ('UndefinedElement' "Data.Array.uninitialized")@)
+-- for initialize new boxed array('Array', 'SmallArray'..).
+--
+uninitialized :: a
+uninitialized = throw (UndefinedElement "Data.Primitive.Array.Class.uninitialized")
+
+-- | A typeclass to unify box & unboxed, mutable & immutable array operations.
+--
+-- Most of these functions simply wrap their primitive counterpart. If there are no primitive ones,
+-- the method is polyfilled using other operations to get the same semantics.
```
cartazio: i dont like the word polyfilled, can we use something like "emulated"?

Data/Primitive/Array/Class.hs
```diff
@@ -0,0 +1,485 @@
...
+--
+-- One exception is that 'shrinkMutableArr' only performs resizing on 'PrimArray' because
+-- the RTS only supports shrinking byte arrays. On other array types, 'shrinkMutableArr' will do nothing.
+--
+class Arr (marr :: * -> * -> *) (arr :: * -> * ) a | arr -> marr, marr -> arr where
```
cartazio: My immediate concern when seeing this interface is that this doesn't allow the same interface to work For ArrayArray too! Then we can't have a nice interface for arrays of unlifted values!

Data/Primitive/UnliftedArray.hs
```diff
@@ -137,6 +141,22 @@ instance PrimUnlifted (MV.MutVar s a) where
   toArrayArray# (MV.MutVar mv#) = unsafeCoerce# mv#
   fromArrayArray# aa# = MV.MutVar (unsafeCoerce# aa#)
 
+instance PrimUnlifted (MVar a) where
+  toArrayArray# (MVar mv#) = unsafeCoerce# mv#
```
cartazio: Wait what. This doesn't seem safe .. could you explain this mapping to me?

Data/Primitive/Array/Class.hs
cartazio: also should this interface also support (when applicable) arrays of unlifted pointer values via ArrayArray?

Data/Primitive/UnliftedArray.hs
cartazio: and all the associated other unsafe coerces too ..

andrewthad: To my understanding, `ArrayArray#` can hold any values whose type is of kind `TYPE UnliftedRep`. This seems safe to me, but it would be nice to have that confirmed by someone else. @bgamari Can you comment on this?

andrewthad:
The current definition of `shrinkMutableArr` for boxed arrays is not good. Even though it's documented, it's semantics are totally wrong. The type signature it has been given preludes a sensible implementation. My vote would be to just remove this method from the typeclass since `resizeMutableArr` can be used to accomplish the same thing.

Data/Primitive/Array/Checked.hs
```diff
@@ -0,0 +1,225 @@
+{-# LANGUAGE TypeFamilies #-}
+
+{-|
+Module      : Data.Array.Checked
+Description : Bounded checked boxed and unboxed arrays
+Copyright   : (c) Winter, 2017
+License     : BSD3
+Maintainer  : libraries@haskell.org, drkoster@qq.com
```
winterland1989: It's my email.

Data/Primitive/Array/Class.hs
winterland1989: This interface support ArrayArray , though the `newArr` for ArrayArray is more unsafe, I should add more doc about this.

Data/Primitive/Array/Checked.hs
winterland1989: I'm not sure because I really need this function inlined to reduce the overhead of bound check.

Data/Primitive/Array/Class.hs
winterland1989: Can you give a reason why that's the case?

Data/Primitive/UnliftedArray.hs
winterland1989: Yes, I think it's safe too, and I'm using this instance in my libuv binding without problems. The newly added instances are really not differ from MutVar , they are unlifted boxed pointers just like ArrayArray.

winterland1989:
@andrewthad I'm agreed that the equal instances is surprising if someone provide a different eq instance for the element. It will not work unless it bit wise equal like all the current primitive type. I  will make the change.

As for the shrinking interface, I don't it is a trouble because ghc never guarantee a shrinking on primitive array is successful, it should be used to optimize  GC of an array: when you are building array with double allocating strategy, you may end with an array with unused element in the tail. You don't want next GC copy them so you want to shrinking the closure size.

I think we can open a issue to ask ghc developers to add shrink for other array types, but this interface should stay IMO.

winterland1989:
I opened a ghc ticket for asking shrink primops for other array type here: https://ghc.haskell.org/trac/ghc/ticket/14124

Data/Primitive/Array/Class.hs
winterland1989: Docs added.

winterland1989:
> Also, for very different reasons, I do not like the overlapping instance for Ord. Overlapping instances mess up instance resolution sometimes. I'll try to come up with an example of this.

@andrewthad  I think i need to be convinced this do have a problem before change `PrimArray Word8` 's `Ord` instance, because `memcmp` is much faster than manually loop throught the word8 array, and i really want this OVERLAPPING instance so that my vector package will not slow down.

AFAICT, `PrimArray Word8` is strictly more specific than `PrimArray a`, and according to GHC manual ghc should be able to choice the right instance.

andrewthad:
@winterland1989 My gripe with `OverlappingInstances` is that it makes it difficult for end-users to work with the instances for a polymorphic `PrimArray`. For example, with your `Ord` instance, this doesn't work:

    myFunc :: (Ord a, Prim a) => PrimArray a -> PrimArray a -> a -> a -> Bool
    myFunc arr1 arr2 val1 val2 = arr1 > arr2 && val1 > val2

You have to turn on `FlexibleContexts` and write this instead:

    myFunc :: (Ord a, Ord (PrimArray a)) => PrimArray a -> PrimArray a -> a -> a -> Bool
    myFunc arr1 arr2 val1 val2 = arr1 > arr2 && val1 > val2

Similarly, things get messed up when you go to derive instances. This doesn't work:

    data NonEmptyPrimArray a = NonEmptyPrimArray a (PrimArray a)
      deriving (Eq,Ord)

You having to use `FlexibleContexts` and `StandaloneDeriving` to do it, and you end up with a `Ord (PrimArray a)` constraint and an `Ord a` constraint as we did in `myFunc`. This is why I do not like `OverlappingInstances`.

The way I see it, there are two ways to write `Eq` and `Ord`. One uses the `Eq`/`Ord` dictionary from the element type and the other just uses `memcmp`. I don't care that much which one gets used, but I would prefer anything to `OverlappingInstances`. So, even if the instances just look like this:

    instance Eq (PrimArray a) where
      (PrimArray baA) == (PrimArray baB) = baA == baB
    
    instance Ord (PrimArray a) where
        (PrimArray baA) `compare` (PrimArray baB) = baA `compare` baB

I would prefer this to `OverlappingInstances`, simply because the constraints that get propagated when working in a polymorphic setting are much easier to think about. The main drawbacks of this approach are that `Ord` now behaves differently depending on the platform's endianness.

Oh, I just thought of a different approach that solves everything and doesn't end up creating weird constraints. What if we added a method to `Prim`:

    class Prim a where
      ...
      compareByteArray# :: a -> ByteArray# -> ByteArray# -> Int#
      compareByteArray# _ arr1 arr2 = ... -- use memcmp

This first argument should always be ignored (as it is for `sizeOf#`), and individual instances can choose to override this if they want to use an `Ord` dictionary instead. This would not cause any breakage downstream because it has a default implementation. The only difficult would be fiddling with the C macros used in `primitive` to generate instances for `Int8`, `Word8`, `Int16`, etc.

andrewthad:
Also, I'm not too concerned about the shrinking stuff. I feel like users of the typeclass could reasonably expect the following law to hold:

    -- this should always evaluate to True
    runST $ do
      arr <- newArr 10
      shrinkMutableArr arr 7
      sz <- sizeofMutableArr marr
      return (sz == 7)

This currently does not hold for instance for boxed arrays. But, I've never actually seen anyone shrink arrays with GHC, so it's not that big of a deal to me, since I don't think it will bother anyone.

winterland1989:
I see your point, fair enough. What about using `SPECIALIZE` rules to rewrite then?

winterland1989:
> Also, I'm not too concerned about the shrinking stuff. I feel like users of the typeclass could reasonably expect the following law to hold:

I have added new documents to explicitly indicate this may NOT hold, in fact on older GHC shrinking `ByteArray` is a no-op too, this will be more correct.  

> This currently does not hold for instance for boxed arrays. But, I've never actually seen anyone shrink arrays with GHC, so it's not that big of a deal to me, since I don't think it will bother anyone.

I'm using it in packing arrays from list, see [here](https://github.com/winterland1989/stdio/blob/master/Data/Vector.hs#L503). The point is that after build an array with double allocating strategy, you may result in an array with unused element, shrinking it will overwrite the heap closure, so that next GC you don't have to copy all the bytes.

andrewthad:
Ah, I didn't even think about specialize rules. Yeah, that would be much easier than anything I suggested.

> I'm using it in packing arrays from list, see here. 

Cool! That's the first time I've actually seen it used. I don't think that `vector` uses it anywhere, although, as you point out, that would be an obvious win for better memory usage.

Data/Primitive/Array/Checked.hs
cartazio: oh,  i meant the inlines everywhere, valid forthis one i guess ;)

Data/Primitive/Array/Class.hs
cartazio: i'm thinking we could have, at least on recent ghc, something like 
`marr :: * -> TYPE r -> *` , and `arr :: TYPE r -> *`, which is like having representation genericity, but in a static / allowed way

Data/Primitive/Array/Class.hs
cartazio: @winterland1989  my point about the term poly filled still stands right?

Data/Primitive/Array/Class.hs
```diff
@@ -75,15 +75,15 @@ uninitialized = throw (UndefinedElement "Data.Primitive.Array.Class.uninitialize
...
--- the RTS only supports shrinking byte arrays. On other array types, 'shrinkMutableArr' will do nothing.
---
 class Arr (marr :: * -> * -> *) (arr :: * -> * ) a | arr -> marr, marr -> arr where
 
     -- | Make a new array with given size.
     --
+    -- It's not safe to access uninitialized element.
     -- For boxed arrays, all elements are @uninitialized@, a thunk that throws
     -- an error if it is forced. For primitive array, elements are uninitialized memory.
+    -- For unlifted array , elements are the refrerence to itself.
```
cartazio: you spelled reference wrong

Data/Primitive/Array/Class.hs
```diff
@@ -140,8 +140,10 @@ class Arr (marr :: * -> * -> *) (arr :: * -> * ) a | arr -> marr, marr -> arr wh
     -- | Resize a mutable array to the given size.
     resizeMutableArr :: (PrimMonad m, PrimState m ~ s) => marr s a -> Int -> m (marr s a)
 
-    -- | Shrink a mutable array to the given size. This operation only works on primitive arrays.
-    -- For boxed arrays, this is a no-op, e.g. 'sizeOfMutableArr' will not change.
+    -- | Shrink a mutable array to the given size.
+    --
+    -- This operation is not guaranteed to have effects, e.g. 'sizeOfMutableArr' will not change.
```
cartazio: i think you could say  something about "this operation is instance dependent, but behavior could be ..."

winterland1989: I don't want to give user an impression of that actually, since i have open an feature request on ghc trac for adding shrink ops to boxed array too, and ben said it's do-able.

Data/Primitive/Array/Class.hs
winterland1989: I'll fix it.

winterland1989:
@cartazio Can you show me when a kind polymorphric array useful? e.g. with `marr :: * -> TYPE r -> *` and `arr :: TYPE r -> *`. I think the box/unboxed instances indicate how the elements are stored pretty clear.

winterland1989:
I have changed overlapping instance to rules now, but the `CPP` marco need `cpphs` to be expanded correctly. will that ba a trouble? If so we can manually expand them.

Data/Primitive/Array/Class.hs
```diff
@@ -0,0 +1,487 @@
...
+    {-# INLINE indexArrM #-}
+    freezeArr = freezeUnliftedArray
+    {-# INLINE freezeArr #-}
+    thawArr = thawUnliftedArray
+    {-# INLINE thawArr #-}
+    unsafeFreezeArr = unsafeFreezeUnliftedArray
+    {-# INLINE unsafeFreezeArr #-}
+    unsafeThawArr (UnliftedArray arr#) = primitive ( \ s0# ->
+            let !(# s1#, marr# #) = unsafeThawArray# (unsafeCoerce# arr#) s0#   -- ArrayArray# and Array# use the same representation
+            in (# s1#, MutableUnliftedArray (unsafeCoerce# marr#) #)            -- so this works
```
winterland1989: I'd like to consult GHC experts here, @bgamari , does my comment hold in this case ? I expect `unsafeThawArray#` works on `ArrayArray#` since they have the same closure type, so this operation will correctly rewrite the closure header. Is that the case?

winterland1989: BTW it's strange that we have `unsafeFreezeArrayArray#` but without a thaw one.

Data/Primitive/UnliftedArray.hs
winterland1989: @bgamari , though i'm pretty sure this works,  I think we need your confirm here too.

andrewthad:
I've PRed a change to [inline the Eq rules]. To me, it has three benefits:

- more readable
- bytewise equality only declared once (instead of once per type)
- you don't have to worry about whatever it was with cpphs that concerned you

andrewthad:
The earlier conversation is burried a little higher in this PR's history, but I want to say that I'm opposed to making `Arr` take levity polymorphic arrays, ie `marr :: * -> TYPE r -> *`. We would have to change the `Prim` typeclass to make this usable. It's not totally to me what the effect on end-user error messages would be.

.travis.yml
```diff
@@ -51,6 +51,7 @@ install:
           $HOME/.cabal/packages/hackage.haskell.org/00-index.tar;
    fi
  - travis_retry cabal update -v
+ - cabal install cpphs
```
cartazio: wait, why do we need cpphs to test this?

primitive.cabal
```diff
@@ -48,8 +51,8 @@ Library
   Build-Depends: base >= 4.5 && < 4.11
                , ghc-prim >= 0.2 && < 0.6
                , transformers >= 0.2 && < 0.6
-
-  Ghc-Options: -O2 -Wall
+  Build-tools: cpphs >= 1.19
+  Ghc-Options: -O2 -Wall -pgmP cpphs -optP --cpp
```
cartazio: why do we need cpphs?

Data/Primitive/PrimArray.hs
```diff
@@ -0,0 +1,470 @@
...
+eqPrimArray paA paB
+    | paA `samePrimArray` paB = True
+    | otherwise =
+        let !sizA = sizeofPrimArray paA
+            !sizB = sizeofPrimArray paB
+        in sizA == sizB && go 0 sizA
+  where
+    go !i !siz | i >= siz  = True
+               | otherwise = indexPrimArray paA i == indexPrimArray paB i && go (i+1) siz
+
```
cartazio: do we have any tests for making sure specialization happens if we look at the compiler verbose output?

Data/Primitive/PrimArray.hs
```diff
@@ -0,0 +1,470 @@
...
+{-# INLINE shrinkMutablePrimArray #-}
+
+-- | Check if the two immutable arrays refer to the same memory block.
+samePrimArray :: PrimArray a -> PrimArray a -> Bool
+{-# INLINE samePrimArray #-}
+samePrimArray (PrimArray (ByteArray ba1#)) (PrimArray (ByteArray ba2#)) =
+    isTrue# (sameMutableByteArray# (unsafeCoerce# ba1#) (unsafeCoerce# ba2#))
+
+--------------------------------------------------------------------------------
+
```
cartazio: lest I forget, we can actually construct analogous prefetch operations for boxed arrays (lifted and unlifted flavors both), using the availaable prefetch operations

Data/Primitive/UnliftedArray.hs
cartazio: i'm not sure i'm comfortable with this code... this is pretty ... dangerous to use, more so than most
https://ghc.haskell.org/trac/ghc/wiki/Commentary/Rts/Storage/HeapObjects documents some of the representation info, i'm not sure if we can in good consciousness expose that in an modulre that doesn't have Unsafe in the name... 

i guess i'm also wondering what we'd use those for?
in some ways it sort of feels like this is a less safe version of the api that the structs lib accomplishes, though @ekmett  and @glguy  can probably opine on this piece perhaps

cartazio:
@andrewthad why cant the class support both lifted and unlifted element arrays for the boxed flavors?

Data/Primitive/Array/Class.hs
cartazio: @winterland1989  can we change polyfilled to emulated with?

andrewthad:
@cartazio Ah, now I see what you're talking about. Yeah, I think that would be an alright small place to try that out.

cartazio:
@andrewthad indeed, and it doesnt' hurt that it would also be even *more generic* than just that, i think?

@winterland1989 @andrewthad i almost feel like this code review would be easier if we just did a long email thread back and forth with explicit diffs off a commit on master? idk.

Data/Primitive/UnliftedArray.hs
dolio: If you look at my `ArrayArray#` wrapping stuff, you'll see the same thing. The primop API is that `ArrayArray#` stores values of type `ArrayArray#`, and you `unsafeCoerce` between that and the type you actually want. I agree it's not a great API, but it's what we have right now.

In the future, maybe there will be an array type that is polymorphic over unlifted types that could be more safe at the primop level. But right now the API I wrote is the safe wrapper around the unsafe API.

Data/Primitive/Array/Class.hs
dolio: I'd be interested to know what the deal with this is, too. Is there some reason that `unsafeThawArrayArray#` doesn't exist?

dolio:
Speaking of levity polymorphism, I kind of wonder off and on how much of this package will be rendered obsolete by it. Wrapping all the primitive array/reference types becomes less necessary, because the boxed, unlifted kind gains most of the advantages of `*` that our wrappers are for. And the type safe `ArrayArray#` wrapper becomes unnecessary, because you can just have a type-parameterized array of unlifted things.

I guess the byte array wrapper still makes sense, and some convenience classes for working with unlifted things can live here. But it might radically alter the package (at least, after we get to a point where versions of GHC without levity polymorphism are considered unworthy of supporting).

.travis.yml
winterland1989: It's for the old eq specialize rules, i have updated to use @andrewthad  's solution now, so this's not needed.

Data/Primitive/PrimArray.hs
winterland1989: I have read the core before, but have got a test yet. Built-in primitive types should not be handled by this code, there're rules to rewrite the loop into `memcmp`.

Data/Primitive/PrimArray.hs
winterland1989: how to do so? by `unsafeCoerce#` the `Array#` into `ByteArray#` ? I'm not sure if this is sensible. let's leave it to future optimizations.

winterland1989:
@cartazio  @dolio   I think the levity polymorphism version may be left to future version, since levity polymorphism is not really happening in base, we don't have to rush for that. Is that OK?

winterland1989:
@cartazio I can make a patch mail to `libraries@haskell.org`, but this patch is rather large and contains several commits, do i have to rebase then? I think github's code review tools is quite nice, but if you prefer email, i can generate one.

And i still can understand your concern with newly added `PrimUnlifted` instances, they works just like how old `MutVar` works, i need this because in one of my project i use an array of `MVar`, i want to cut the memory overhead.

cartazio:
Hrmmm. Are we talking about MVar / or MVar#?

I guess I'm just confused cause I'm learning something new. Care to point
me to the way you'd use this?

winterland1989:
I'm talking `UnliftedArray (MVar a)` which actually store `MVar#` s. the use case is here: https://github.com/winterland1989/stdio/blob/master/System/IO/UV/Manager.hs#L67 

In short words: I need an array of `MVar` to block the thread doing read/write. and after read/write finish i unblock them by index.

cartazio:
ok, so the unsafe coerce shouldn't be needed in ghc 8.2 right because you
can just unbox the mvar's directly into the correct kind, right?

andrewthad:
@cartazio the unsafe coerce is still needed in GHC 8.2. To do what you're describing, we would need a primitive like this:

    writeArrayArrayUnlifted# :: forall (a :: TYPE UnliftedRep). MutableArrayArray# s -> Int# -> a -> State# s -> State# s

Such a function could replace `writeArrayArray#`, `writeByteArrayArray#`, `writeMutableByteArrayArray#`. However, we don't have a primitive that does this, even in 8.2.

winterland1989:
@cartazio It's irrelevant to levity polymorphism , the trick's idea is that `ArrayArray#` is an unlifted boxed type, which `MutVar#`, `MVar#` .. can be cast into and from without runtime danger. The `unsafeCoerce#` bit is doing the type cast, not kind.

Data/Primitive/UnliftedArray.hs
winterland1989: @dolio  Yes, and i extend your trick to many more unlifted boxed type wrapper, such as `IORef/STRef` and `MVar/TVar`, i think @cartazio didn't get this at first, maybe your explain will make him feel ensured.

winterland1989:
@cartazio  I wrote a blog post about the unlifted array stuff: http://winterland.me/2017/08/18/an%20unified%20array%20interface/ . I hope that can clear your concern about my extension to `PrimUnlifted`.

BTW, this patch is ready, please consider merging if no other problems.

cartazio:
Awesome. I'll do a final go over later today / tomorrow

Data/Primitive/ByteArray.hs
```diff
@@ -262,3 +403,79 @@ instance Typeable s => Data (MutableByteArray s) where
...
+isMutableByteArrayPinned :: MutableByteArray s -> Bool
+{-# INLINE isMutableByteArrayPinned #-}
+isMutableByteArrayPinned (MutableByteArray mba#) =
+    c_is_mutable_byte_array_pinned mba# > 0
+
+foreign import ccall unsafe "hsprimitive_is_byte_array_pinned"
+    c_is_byte_array_pinned :: ByteArray# -> CInt
+
+foreign import ccall unsafe "hsprimitive_is_byte_array_pinned"
+    c_is_mutable_byte_array_pinned :: MutableByteArray# s -> CInt
```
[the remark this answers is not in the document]
winterland1989: Yes, you're right. I've add CPPs to use primops where possible.

dolio:
Sorry, I've been a bit distracted lately. I'm going to look through this a bit closer in preparation for merging (also see what conflicts need to be solved; probably nothing major, so I can maybe do it myself, without needing you to rebase).

.travis.yml
dolio: If this isn't needed anymore, can we remove it? :)

winterland1989:
@dolio  Thanks, if need anything please let me know.

Data/Primitive/Array/Class.hs
```diff
@@ -0,0 +1,497 @@
...
+--
+uninitialized :: a
+uninitialized = throw (UndefinedElement "Data.Primitive.Array.Class.uninitialized")
+
+-- | A typeclass to unify box & unboxed, mutable & immutable array operations.
+--
+-- Most of these functions simply wrap their primitive counterpart. If there are no primitive ones,
+-- the method is emulated using other operations to get the same semantics.
+--
+class Arr (marr :: * -> * -> *) (arr :: * -> * ) a | arr -> marr, marr -> arr where
```
treeowl: I don't understand why everything is lumped in one class. I think it makes more sense to have three:

```haskell
class MutArr (marr :: * -> * -> *) a where
  newArr :: ...
  newArrWith :: ...
  readArr :: ...
  writeArr :: ...
  setArr :: ...
  copyMutableArr :: ...
  moveArr :: ...
  cloneMutableArr :: ...
  sameMutableArr :: ...
  sizeofMutableArr :: ...
  shrinkMutableArr :: ...
  resizeMutableArr :: ...

class ImmutArr (arr :: * -> *) a where
  indexArr# :: arr a -> Int -> (# a #)
  -- Leave indexArr and indexArrM out of the class as they can easily be implemented
  -- as functions given indexArr#; going the other way is a good bit harder for humans
  -- and optimizers.

  -- cloneArr strikes me as a bit fishy. What's its story?
  cloneArr :: ...
  sameArr :: ...
  sizeofArr :: ...

class (ImmutArr arr a, MutArr marr a) => Arr marr arr a | arr -> marr, marr -> arr where
  freezeArr :: ...
  unsafeFreezeArr :: ...
  thawArr :: ...
  unsafeThawArr :: ...
  copyArr :: ...
```

Data/Primitive/Array/Class.hs
```diff
@@ -0,0 +1,497 @@
...
+    {-# INLINE copyMutableArr #-}
+
+    moveArr marr1 s1 marr2 s2 l
+        | l <= 0 = return ()
+        | sameMutableArray marr1 marr2 =
+            case compare s1 s2 of
+                LT ->
+                    let !d = s2 - s1
+                        !s2l = s2 + l
+                        go !i | i >= s2l = return ()
```
treeowl: Use `when` or `unless` to avoid writing out `return ()` branches.

Data/Primitive/PrimArray.hs
```diff
@@ -0,0 +1,472 @@
...
+{-# RULES "eqPrimArray/Word32" eqPrimArray = memcmpPrimArray (undefined :: Word32) #-}
+{-# RULES "eqPrimArray/Word64" eqPrimArray = memcmpPrimArray (undefined :: Word64) #-}
+{-# RULES "eqPrimArray/Int8"   eqPrimArray = memcmpPrimArray (undefined :: Int8) #-}
+{-# RULES "eqPrimArray/Int16"  eqPrimArray = memcmpPrimArray (undefined :: Int16) #-}
+{-# RULES "eqPrimArray/Int32"  eqPrimArray = memcmpPrimArray (undefined :: Int32) #-}
+{-# RULES "eqPrimArray/Int64"  eqPrimArray = memcmpPrimArray (undefined :: Int64) #-}
+
+{-# RULES "eqPrimArray/Addr"   eqPrimArray = memcmpPrimArray (undefined :: Addr) #-}
+{-# RULES "eqPrimArray/Ptr"    eqPrimArray = memcmpPrimArray (undefined :: Ptr a) #-}
+{-# RULES "eqPrimArray/FunPtr" eqPrimArray = memcmpPrimArray (undefined :: FunPtr a) #-}
```
treeowl: Ugh... Wouldn't it be better to use a class for `eqPrimArray`? Rules are fragile, and require user involvement, and so on. It's not great. Separately, passing `undefined` is rather old-fashioned. Use a proxy type instead.

Data/Primitive/PrimArray.hs
cartazio: i'll revisit this after i do a proper code review

cartazio:
i need to revisit this PR and review it properly

andrewthad:
Currently, https://github.com/haskell/primitive/pull/106 implements just the `PrimArray` part of this PR. The other main parts of this are:

1. A typeclass unifying the interface to arrays
2. A (slower) bounds-checking interface to array that can easily be switched in while debugging
3. Function for checking if byte array is pinned
4. Prefetching for byte arrays
5. Additional `PrimUnlifted` instances

Some of these have been addressed by recent PRs. I suspect that, at this point, it may be best to break pieces off of this PR into other PRs. This is especially important for typeclass that unifies the interface to arrays, since there's some bikeshedding possible there.

chessai:
After reading this over, I much prefer the API defined in `contiguous`: https://hackage.haskell.org/package/contiguous

Maybe others can chime in here.

chessai:
Also, the PrimArray changes have already happened. I'm referring specifically to the `Arr` typeclass.

andrewthad:
The `PrimArray` type has been added to `primitive`, and this PR has bitrotted in the last three years. I've going to close this, but if anyone wants to discuss adding an `Arr` typeclass to `primitive`, open an new issue to discuss this.

PR #66, opened 2017-09-13
Add sconcat implementations to Semigroup instances

bgamari:
As of GHC 8.4 Semigroup is a superclass of Monoid. Account for this
where necessary.

### DIFF
--- a/Data/Primitive/Array.hs
+++ b/Data/Primitive/Array.hs
@@ -34,6 +34,7 @@ import qualified GHC.Exts as Exts
 import GHC.Exts (fromListN, fromList)
 #endif
 
+import qualified Data.Foldable as F
 import Data.Typeable ( Typeable )
 import Data.Data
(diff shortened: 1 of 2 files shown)


RyanGlScott:
I did add `Semigroup` instances in 7e6e7b4667020b61986c60e0c2d642e700e7966d, but I didn't add `sconcat` implementations. I suppose we ought to.

bgamari:
I've rebased this patch, which now merely adds `sconcat` implementations.

Data/Primitive/Array.hs
```diff
@@ -34,6 +34,7 @@ import qualified GHC.Exts as Exts
 import GHC.Exts (fromListN, fromList)
 #endif
 
+import qualified Data.Foldable as F
```
RyanGlScott: This import is only used within the scope of an `#if MIN_VERSION_base(4,9,0) ... #endif` block, so on old GHCs this is an unused import warning. I think moving the import next to `Data.Semigroup` (which is appropriately guarded by CPP) should fix this.

PR #68, opened 2017-10-13
Mention unconventional nature of ByteArray's Ord instance

bgamari:


glguy:
Would it be worth explaining in the documentation that the lengths are considered first? If that's not intended to be a guaranteed behavior of the API could we be explicit that no guarantees are offered?

bgamari:
Eric Mertens <notifications@github.com> writes:

> Would it be worth explaining in the documentation that the lengths are
> considered first? If that's not intended to be a guaranteed behavior
> of the API could we be explicit that no guarantees are offered?
>
Indeed I intentionally wrote the documentation to not state the precise
ordering used. However, I can make the lack of guarantees more explicit.

RyanGlScott:
@bgamari, what is the status of this?

Issue #76, opened 2018-02-27
Semigroup and Monoid instances for ByteArray


andrewthad:
Can I add `Semigroup` and `Monoid` instances for `ByteArray`?

cartazio:
Concating or some operation on the bytes?

andrewthad:
Sorry I didn't clarify that. Yes, concatenation. Zipping them together with conjunction or disjunction or xor or anything would require an infinitely long bytearray as `mempty`, so a `Monoid` instance wouldn't even be possible.

cartazio:
Does it require an infinite mempty?

So there’s several legal semi group instances and only the concatenation
one is also a monoid

andrewthad:
Yes, there are several possible semigroup instances. I think that concatenating is the most reasonable choice because:

- The `Monoid` instance for `Array` is defined with concatenation as `mappend`.
- No other semigroup instance gives rise to a `Monoid`.

cartazio:
but ... is that the right way to support concatenation? It might be better
to provide a ByteArray Builder, so that repeated concat has the right
complexity, and provide the monoid instance for that instead, i think

andrewthad:
But `primitive` isn't really an appropriate place for a builder. All other analogous types have this same concatenating `Monoid` instance: `Array`, `ByteString`, `Vector` (including primitive vectors). It's seldom a performant way to do things, but it's often convenient. Also, `mconcat` is useful here and can be overridden with something that gives much better complexity than its default implementation.

cartazio:
perhaps you're right, what better complexity strategy do you have in mind?

cartazio:
oh gotchs, good point

andrewthad:
I've implemented this in https://github.com/haskell/primitive/pull/77

cartazio:
@dolio @RyanGlScott @hvr  thoughts on this?

RyanGlScott:
I'll comment on the pull request.

RyanGlScott:
#77 has been merged.

PR #77, opened 2018-03-05
add Semigroup and Monoid instances for ByteArray

andrewthad:
Implements Semigroup and Monoid instances for ByteArray described in https://github.com/haskell/primitive/issues/76

This code has not yet been tested. The functions `sconcat`, `stimes`, and `mconcat` are given implementations that are asymptotically superior to the default implementations.

### DIFF
--- a/Data/Primitive/ByteArray.hs
+++ b/Data/Primitive/ByteArray.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE CPP, MagicHash, UnboxedTuples, UnliftedFFITypes, DeriveDataTypeable #-}
+{-# LANGUAGE BangPatterns, CPP, MagicHash, UnboxedTuples, UnliftedFFITypes, DeriveDataTypeable #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeFamilies #-}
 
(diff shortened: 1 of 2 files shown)


RyanGlScott:
Well, I'll state the obvious: we should add tests for this! Currently, the [test suite](https://github.com/haskell/primitive/tree/53f72ce69a4dfde5345cf5809a8b4a1993523367/test) is rather primitive (har har), but adding some unit tests for things like mappending empty `ByteArray`s, the associativity of `mappend`, etc. should be within reach.

andrewthad:
har har indeed :) I'll add some tests for this.

andrewthad:
I've added some passing tests

RyanGlScott:
Assuming Travis validates, LGTM.

RyanGlScott:
It looks like older GHCs are failing:

```
Data/Primitive/ByteArray.hs:393:10:
    Not in scope: type constructor or class `Monoid'
    Perhaps you meant `Monad' (imported from Prelude)
```

This might be fixed with by adding the import:

```haskell
#if !(MIN_VERSION_base(4,8,0))
import Data.Monoid (Monoid(..))
#endif
```

Also, you're likely going to need to CPP out the `Semigroup`–specific stuff in the test suite, since that's only been in `base` since 4.9.0.

RyanGlScott:
Looking better. I think if you guarded the import of `Data.Semigroup` in the test suite with `#if MIN_VERSION_base(4,9,0) ... #endif`, that'd do the trick.

andrewthad:
Perfect, thanks! Passing with everything now.

PR #81, opened 2018-03-14
ignore build artifacts from cabal new-build and stack

andrewthad:


RyanGlScott:
Yikes, I'm shocked that this hasn't been done before.

While you're at it, do you mind just using the template Haskell `.gitignore` [here](https://github.com/github/gitignore/blob/3edcb9bd7126356721db8969549d8c6863b5dfb0/Haskell.gitignore)? That should save us some huffing and puffing down the line.

andrewthad:
I've copied in that one and added `stack.yaml` to it.

Issue #80, opened 2018-03-14
Copying between address and mutable byte array


andrewthad:
There are several functions in `GHC.Prim` for copying between an `Addr#` and a `MutableByteArray#`. For example:

    copyAddrToByteArray# :: Addr# -> MutableByteArray# s -> Int# -> Int# -> State# s -> State# s

I'd like to add wrappers for these to `primitive`. I'd document the minimum version of GHC required for these function to be available since I'm pretty sure these haven't been available forever, and `primitive` seems to be very backwards compatible. Would it be better to put them in `Data.Primitive.ByteArray` or `Data.Primitive.Addr`?

cartazio:
Sounds reasonable.  Not sure which spot would be best.  Perhaps Ryan or Dan
have opinions

andrewthad:
For reference:

    copyByteArrayToAddr# :: ByteArray# -> Int# -> Addr# -> Int# -> State# s -> State# s
    copyMutableByteArrayToAddr# :: MutableByteArray# s -> Int# -> Addr# -> Int# -> State# s -> State# s
    copyAddrToByteArray# :: Addr# -> MutableByteArray# s -> Int# -> Int# -> State# s -> State# s

I'm planning on putting the first two into `Data.Primitive.ByteArray` and the last one in `Data.Primitive.Addr`.

PR #82, opened 2018-03-14
Copy bytearray addr

andrewthad:
Resolves https://github.com/haskell/primitive/issues/80

Currently, this does not provide these functions for users of older GHCs. I'm not sure if this how important this is. Also, I'm not sure when the primitives for copying between a byte array and an address were introduced to `ghc-prim`. I guessed GHC 7.2, but hopefully travis will let me know if I guessed wrong.

### DIFF
--- a/Data/Primitive/Addr.hs
+++ b/Data/Primitive/Addr.hs
@@ -31,7 +31,9 @@ module Data.Primitive.Addr (
 
 import Control.Monad.Primitive
 import Data.Primitive.Types
+#if __GLASGOW_HASKELL__ >= 708
 import Data.Primitive.ByteArray
+#endif
 
 import GHC.Base ( Int(..) )
 import GHC.Prim
--- a/Data/Primitive/Addr.hs
+++ b/Data/Primitive/Addr.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE MagicHash, UnboxedTuples #-}
+{-# LANGUAGE MagicHash, UnboxedTuples, CPP #-}
 
 -- |
 -- Module      : Data.Primitive.Addr
@@ -22,11 +22,16 @@ module Data.Primitive.Addr (
   indexOffAddr, readOffAddr, writeOffAddr,
 
   -- * Block operations
-  copyAddr, moveAddr, setAddr
+  copyAddr,
+#if __GLASGOW_HASKELL__ >= 702
+  copyAddrToByteArray,
+#endif
+  moveAddr, setAddr
 ) where
 
 import Control.Monad.Primitive
 import Data.Primitive.Types
+import Data.Primitive.ByteArray
 
 import GHC.Base ( Int(..) )
 import GHC.Prim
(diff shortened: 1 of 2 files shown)


andrewthad:
This is now fixed

RyanGlScott:
Looking good.

One suggestion: note in the Haddocks for these functions are only available on GHC 7.8 and later.

hvr:
To avoid undesirable error-prone conditional APIs, there's a different encoding that I've been considering for `unix`:

Instead of

```
copyByteArrayToAddr :: PrimMonad m  => Addr -> ByteArray -> Int  -> Int  -> m ()
```

use e.g.

```
copyByteArrayToAddr :: PrimMonad m  => Maybe (Addr -> ByteArray -> Int  -> Int  -> m ())
```

RyanGlScott:
While the sentiment is appreciated, I don't think it's worthwhile to adopt this approach here. This is the `primitive` library after all—it's understood that if you're using functions from this library, it's your responsibility to ensure that you use them correctly. Part of that is reading the documentation, so I think a simple disclaimer in the Haddocks should suffice.

andrewthad:
@hvr If these were operations that weren't actually available on every platform, I would agree. But, these are actually available everywhere. In fact, they could be shimmed out with an FFI call like they are for `copyByteArray`, `copyMutableByteArray`, etc. on old versions of GHC. I didn't try doing this because I don't actually need to use the functions on GHCs older than 7.8, but it should be possible. It would just need to be tested.

andrewthad:
I've added the disclaimer.

Data/Primitive/Addr.hs
```diff
@@ -22,11 +22,16 @@ module Data.Primitive.Addr (
...
+  copyAddr,
+#if __GLASGOW_HASKELL__ >= 708
+  copyAddrToByteArray,
+#endif
+  moveAddr, setAddr
 ) where
 
 import Control.Monad.Primitive
 import Data.Primitive.Types
+import Data.Primitive.ByteArray
```
RyanGlScott: Guard this import with `#if __GLASGOW_HASKELL__ >= 708` too. Otherwise, it'll be redundant on older GHCs.

Data/Primitive/ByteArray.hs
```diff
@@ -231,6 +235,39 @@ copyMutableByteArray (MutableByteArray dst#) doff
...
+  -> ByteArray -- ^ source array
+  -> Int -- ^ offset into source array
+  -> Int -- ^ number of bytes to copy
+  -> m ()
+{-# INLINE copyByteArrayToAddr #-}
+copyByteArrayToAddr (Addr dst#) (ByteArray src#) soff sz
+  = primitive_ (copyByteArrayToAddr# src# (unI# soff) dst# (unI# sz))
+#endif
+
+#if __GLASGOW_HASKELL__ >= 708
```
RyanGlScott: Why not delete this `#if __GLASGOW_HASKELL__ >= 708` and put `copyMutableByteArrayToAddr` within the same `#if __GLASGOW_HASKELL__ >= 708` block as `copyByteArrayToAddr`?

hvr:
> If these were operations that weren't actually available on every platform

are they available on GHCJS or Eta?

andrewthad:
They should be. If they are not, I would consider that a bug with GHCJS/Eta. They were both forked from GHC after the primitives in question were added.

Data/Primitive/ByteArray.hs
andrewthad: done

Data/Primitive/Addr.hs
andrewthad: done

andrewthad:
Also, on a more fundamental level, I think that the difference between `unix` and `primitive` concerning the Maybe-vs-leave-it-out question is that in `primitive`, the operations really aren't supposed to be missing on any platform. If they are, it's a mistake, and there is always a way to shim them (although it may be difficult, I have no idea how the Eta FFI looks). With `unix`, there are some distros that don't fully conform to posix, and the functions cannot necessarily be shimmed.

hvr:
> If they are not, I would consider that a bug with GHCJS/Eta. They were both forked from GHC after the primitives in question were added.

What if a target doesn't support the concept of an `Addr#`?

andrewthad:
Then half of the `Prim` typeclass and the entire `Data.Primitive.Addr` module would be broken. I wouldn't consider anything that doesn't support `Addr#` to be a fork of GHC. There's a bunch of stuff in `base` that wouldn't compile. Everything dealing with `Ptr` and `ForeignPtr` would be broken. Not only would such a target not be a fork of GHC, it's unlikely that it could be considered Haskell at all since the language report mandates the existence of `Ptr`. If a target doesn't support, as you say, the "concept of an `Addr#`", how could it support the concept of a `Ptr`?

RyanGlScott:
Looks good to me. I'll go ahead and merge, and we can decide later if it's worth the effort to try and backport these functions on older versions of GHC.

PR #87, opened 2018-03-18
Fold better; fix foldl1 bug

### LINKED ISSUE #86: foldl1 looks flat-out wrong
I'm working on a patch that will replace the folds anyway, but the current code,

```haskell
  foldl1 f a | sz == 0   = die "foldl1" "empty array"
             | otherwise = go $ sz-1
   where sz = sizeofArray a
         z = indexArray a 0
         go i | i < 1     = f (go $ i-1) (indexArray a i)
              | otherwise = z
```

doesn't look remotely right. The `i < 1` should be `i > 0` or (probably better) `i /= 0`, or the branches should be swapped, or something.

treeowl:
* Index `Array`s and `SmallArray`s eagerly.
* Fix `foldl1` for `Array`, which was completely wrong.
* Make the structure of the `Foldable` instances for `Array`
  and `SmallArray` the same, taking what I thought was the best
  from each.

Fixes #86

### DIFF
--- a/Data/Primitive/SmallArray.hs
+++ b/Data/Primitive/SmallArray.hs
@@ -6,6 +6,7 @@
 {-# LANGUAGE DeriveTraversable #-}
 {-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE BangPatterns #-}
 
 -- |
 -- Module : Data.Primitive.SmallArray
@@ -225,6 +226,15 @@ indexSmallArray (SmallArray a) = indexArray a
 #endif
 {-# INLINE indexSmallArray #-}
 
+#if HAVE_SMALL_ARRAY
+-- | Read a value from the immutable array at the given index, returning
+-- the result in an unboxed unary tuple. This is currently used to implement
+-- folds.
+indexSmallArray## :: SmallArray a -> Int -> (# a #)
+indexSmallArray## (SmallArray ary) (I# i) = indexSmallArray# ary i
+{-# INLINE indexSmallArray## #-}
+#endif
+
 -- | Create a copy of a slice of an immutable array.
 cloneSmallArray
   :: SmallArray a -- ^ source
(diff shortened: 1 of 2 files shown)


cartazio:
Look pretty solid.

Could you explain what you mean by structure of the foldable instances ?
The code looks clean and good, I’m just not sure if you mean formatting of
instances or semantics :)

treeowl:
Oh, I just meant the way the code is written. `fix` vs explicit recursion, the way the first/last element is handled in `foldr1`/`foldl`. I also removed the explicit `toList` implementations because they really just did the same as the default ones in the same way, which seemed a bit silly.

Data/Primitive/SmallArray.hs
```diff
@@ -424,53 +434,96 @@ instance Ord a => Ord (SmallArray a) where
...
+  minimum ary | sz == 0   = die "minimum" "Empty SmallArray"
+              | (# frst #) <- indexSmallArray## ary 0
+              = go 1 frst
+   where sz = sizeofSmallArray ary
+         go i !e
+           | i == sz = e
+           | (# x #) <- indexSmallArray## ary i
+           = go (i+1) (min e x)
+  {-# INLINE minimum #-}
+  sum = foldl' (+) 0
```
RyanGlScott: Out of curiosity, does this generate better code than the default implementation for `sum` (which is something like `getSum . foldMap Sum`)?

treeowl:
I think we can merge this now. I don't want to have to keep rebasing the
traversal stuff over it.

treeowl:
Thanks.

andrewthad:
For posterity's sake, I'll point out that `foldl1` is incorrect in this PR but has been fixed in https://github.com/haskell/primitive/pull/99.

treeowl:
I'm surprised. I did test it manually with the list monoid. Maybe I did
something stupid afterwards?

andrewthad:
Maybe you accidentally picked a list that was the reversal of itself. Or maybe you accidentally revert part of your changes before committing. Either way, not a big deal now that we have decent tests.

PR #88, opened 2018-03-18
Traverse better

### LINKED ISSUE #85: Consider a different traverse implementation for Array
Some informal benchmarks suggest it's better to traverse by folding up something like

```haskell
newtype STA a = STA (forall s. MutableArray# s a -> ST s (Array a))
```

in the functor, and then mapping over to initialize. Also, traversing in `PrimBase` functors is much more direct. It might be worth adding rewrite rules rewriting `traverse` in `ST` and `IO` to the special forms they support.

treeowl:
* Instead of traversing a list and then converting to an array,
be more direct. GHC seems to be able to optimize this better,
at least in common cases.

* Add functions for traversing arrays in "affine" primitive monads: monad transformer stacks built from `StateT`, `ReaderT`, `WriterT`, `RWST`, `ExceptT`, `Stream (Of a)`, etc., but not, for example, `ListT` (done right or otherwise) or `LogicT`.

Fixes #85.

### DIFF
--- a/Data/Primitive/Array.hs
+++ b/Data/Primitive/Array.hs
@@ -22,7 +22,8 @@ module Data.Primitive.Array (
   copyArray, copyMutableArray,
   cloneArray, cloneMutableArray,
   sizeofArray, sizeofMutableArray,
-  fromListN, fromList
+  fromListN, fromList,
+  unsafeTraverseArray
 ) where
 
 import Control.Monad.Primitive
@@ -55,6 +56,9 @@ import Data.Monoid
 #if MIN_VERSION_base(4,9,0)
 import qualified Data.Foldable as F
 import Data.Semigroup
+#endif
+#if MIN_VERSION_base(4,8,0)
+import Data.Functor.Identity
 #endif
 
 import Text.ParserCombinators.ReadP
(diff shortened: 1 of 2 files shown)


treeowl:
Do we still support pre-7.2 GHC, or can I kill that CPP?

cartazio:
We don’t support pre 7.2, and in fact I think because 7.2 was just a tech
preview release we may only currently support 7.4 onwards ?   I could be
wrong though

treeowl:
I amended the commit to avoid adding 7.2-related CPP, and added a commit to remove the rest of said CPP from `Data.Primitive.Array`.

treeowl:
Hrm..... Turned out what I was doing to turn the `letrec` to a `joinrec` in the identity case was a bit strictness-risky in the general case. I'm not sure why the join point stuff is fragile here. But I decided to just do the simple thing for now. We can revisit it some other time.

treeowl:
Actually, I think I see how to fix that. It's a bit ugly, and only available for recent GHC, but I think I'll do it anyway. One more fix incoming...

treeowl:
(Ugly as in ugly source code. It's actually cleaner in a sense.

treeowl:
Oh, actually, I can make it work for older versions too, with a perfectly safe `unsafeCoerce#`. Much less mess that way.

treeowl:
Eh, I changed my mind. Source code gets too messy, and the benefits seem too limited. I think this is ready to merge as is.

Issue #89, opened 2018-03-19
Add a proper test suite


treeowl:
I just went to try to add QuickCheck or SmallCheck properties to the test suite, and discovered that there isn't a proper test suite. Ack! No wonder the `Foldable` instance for `Array` managed to be wrong for so long! I'll be happy to write some properties, but I don't have the time right now to set up the framework.

RyanGlScott:
Alas, this is not as easy to accomplish as it sounds. `QuickCheck` in particular depends on `tf-random`, which in turn depends on `primitive` itself, so you quickly run into https://github.com/haskell/cabal/issues/5200 if you try to make `primitive`'s test suite depend on it.

That being said, there is a relatively serviceable workaround: turn the test suite into its own `.cabal` file and make it its own project, distinct from `primitive` itself. That way, build tools can first build `primitive` and then build `tf-random` _et al._ using the locally built `primitive`, avoiding any dependency cycles. I've used this hack before to make a [test suite for `transformers-compat`](https://github.com/ekmett/transformers-compat/pull/33), and it seems to work okay.

RyanGlScott:
I've opened #90 to implement the suggestion in https://github.com/haskell/primitive/issues/89#issuecomment-374210328. Once that is merged (pending Travis giving a green check mark), that should make setting up `QuickCheck` tests far, far easier.

treeowl:
Great! Have you had a chance to look at my folds and traversals?

On Mar 19, 2018 10:02 AM, "Ryan Scott" <notifications@github.com> wrote:

I've opened #90 <https://github.com/haskell/primitive/pull/90> to implement
the suggestion in #89 (comment)
<https://github.com/haskell/primitive/issues/89#issuecomment-374210328>.
Once that is merged (pending Travis giving a green check mark), that should
make setting up QuickCheck tests far, far easier.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/haskell/primitive/issues/89#issuecomment-374222243>,
or mute
the thread

andrewthad:
@RyanGlScott Thanks for moving the test suite into its own project. I'm going to work on adding some property tests.

treeowl:
Andrew, please let me know if/when you start, and with what. I'll do the
same.

andrewthad:
I'm working on it now. I'm going to use `tasty` + `quickcheck` + `quickcheck-classes` just as an easy way to get access to some of the typeclass property tests we're interested in.

[...]

PR #90, opened 2018-03-19
Move tests to separate .cabal file 

RyanGlScott:
This is a preparatory refactor towards #89. This moves the test suite into a separate `.cabal` file so that in the future, we can modify the test suite to depend on libraries that depend on `primitive` itself, such as `QuickCheck`.

Note that this does not change any of the code in the test suite itself—let's save that for a separate commit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -93,15 +93,16 @@ install:
       cabal new-update head.hackage -v
     fi
   - grep -Ev -- '^\s*--' ${HOME}/.cabal/config | grep -Ev '^\s*$'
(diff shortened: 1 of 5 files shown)


treeowl:
This sounds like an approach that could also be helpful for `containers` and `unordered-containers`.

cartazio:
Woah.  Are we adjusting CI TO ensure they are still run?!

RyanGlScott:
@cartazio, the tests are most certainly still being run (see https://travis-ci.org/haskell/primitive/jobs/355374067#L733 as evidence of this).

PR #93, opened 2018-03-19
Fix fmap for Array

### LINKED ISSUE #92: The Functor instance for Array is completely wrong
Currently,

```haskell
instance Functor Array where
  fmap f a =
    createArray (sizeofArray a) (die "fmap" "impossible") $ \mb ->
      let go i | i < sizeofArray a = return ()
               | otherwise         = writeArray mb i (f $ indexArray a i)
                                  >> go (i+1)
       in go 0
```

That should be

```haskell
   | i == sizeofArray a = return ()
```

treeowl:
`fmap` for `Array` got its indexing all wrong. Fix that. Fix an infinite loop in `>>=` for `Array`.

Fixes #92 and #95

### DIFF
--- a/Data/Primitive/Array.hs
+++ b/Data/Primitive/Array.hs
@@ -474,9 +474,9 @@ fromList l = fromListN (length l) l
 instance Functor Array where
   fmap f a =
     createArray (sizeofArray a) (die "fmap" "impossible") $ \mb ->
-      let go i | i < sizeofArray a = return ()
-               | otherwise         = writeArray mb i (f $ indexArray a i)
-                                  >> go (i+1)
+      let go i | i == sizeofArray a = return ()
+               | otherwise          = writeArray mb i (f $ indexArray a i)
+                                      >> go (i+1)
        in go 0
 #if MIN_VERSION_base(4,8,0)
   e <$ a = runST $ newArray (sizeofArray a) e >>= unsafeFreezeArray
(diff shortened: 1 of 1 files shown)


Data/Primitive/Array.hs
```diff
@@ -474,9 +474,9 @@ fromList l = fromListN (length l) l
 instance Functor Array where
   fmap f a =
     createArray (sizeofArray a) (die "fmap" "impossible") $ \mb ->
-      let go i | i < sizeofArray a = return ()
-               | otherwise         = writeArray mb i (f $ indexArray a i)
-                                  >> go (i+1)
+      let go i | i == sizeofArray a = return ()
```
cartazio: yeah... good catch, looks like this was never tested or used ever before

Data/Primitive/Array.hs
```diff
@@ -521,17 +521,20 @@ instance Alternative Array where
...
-
-   build sz stk = createArray sz (die ">>=" "impossible") $ \mb ->
-     let go off (b:bs) = copyArray mb off b 0 (sizeofArray b) >> go (off + sizeofArray b) bs
-         go _   [    ] = return ()
-      in go 0 stk
+   la = sizeofArray ary
+   collect sz stk i
+     | i < 0 = createArray sz (die ">>=" "impossible") $ fill 0 stk
+     | otherwise = let sb = f $ indexArray ary i in
+         collect (sz + sizeofArray sb) (sb:stk) (i-1)
```
RyanGlScott: Is the key part of fixing the infinite loop the use of a minus sign here instead of a plus sign? (It's not obvious at a quick glance.)

treeowl: I never really dug into that bit. The `SmallArray` code seemed to work, do I copied it.

PR #99, opened 2018-03-20
initial implementation of a more comprehensive test suite

andrewthad:
Still a work in progress

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,6 +74,7 @@ install:
   - HADDOCK=${HADDOCK-true}
   - INSTALLED=${INSTALLED-true}
   - GHCHEAD=${GHCHEAD-false}
+  - CABALFLAGS="--flags='quickcheck-classes -aeson -semigroupoids'"
   - travis_retry cabal update -v
   - "sed -i.bak 's/^jobs:/-- jobs:/' ${HOME}/.cabal/config"
   - rm -fv cabal.project cabal.project.local
@@ -123,7 +124,7 @@ script:
   - if $INSTALLED; then echo cabal new-build -w ${HC} --disable-tests --disable-benchmarks $(${HCPKG} list --global --simple-output --names-only | sed 's/\([a-zA-Z0-9-]\{1,\}\) */--constraint="\1 installed" /g') all | sh; else echo "Not building with installed constraints"; fi
 
   # build & run tests, build benchmarks
-  - cabal new-build -w ${HC} ${TEST} ${BENCH} -j2 all
+  - cabal new-build -w ${HC} ${CABALFLAGS} ${TEST} ${BENCH} -j2 all
   - if [ "x$TEST" = "x--enable-tests" ]; then cabal new-test -w ${HC} ${TEST} ${BENCH} all; fi
 
   # cabal check
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,7 +74,7 @@ install:
   - HADDOCK=${HADDOCK-true}
   - INSTALLED=${INSTALLED-true}
   - GHCHEAD=${GHCHEAD-false}
-  - CABALFLAGS=--constraint='quickcheck-classes -aeson -semigroupoids'
+  - CABALFLAGS="--flags=\"quickcheck-classes -aeson -semigroupoids\""
   - travis_retry cabal update -v
   - "sed -i.bak 's/^jobs:/-- jobs:/' ${HOME}/.cabal/config"
   - rm -fv cabal.project cabal.project.local
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,7 +74,7 @@ install:
   - HADDOCK=${HADDOCK-true}
   - INSTALLED=${INSTALLED-true}
   - GHCHEAD=${GHCHEAD-false}
-  - CABALFLAGS="--flags='quickcheck-classes -aeson -semigroupoids'"
+  - CABALFLAGS=--flags='quickcheck-classes -aeson -semigroupoids'
   - travis_retry cabal update -v
   - "sed -i.bak 's/^jobs:/-- jobs:/' ${HOME}/.cabal/config"
   - rm -fv cabal.project cabal.project.local
--- a/.travis.yml
+++ b/.travis.yml
@@ -118,14 +118,14 @@ script:
   - "printf 'packages: primitive-*/*.cabal primitive-tests-*/*.cabal\\n' > cabal.project"
   - cat cabal.project
   # this builds all libraries and executables (without tests/benchmarks)
-  - cabal new-build -w ${HC} --disable-tests --disable-benchmarks all
+  - cabal new-build -w ${HC} ${CABALFLAGS} --disable-tests --disable-benchmarks all
 
   # Build with installed constraints for packages in global-db
-  - if $INSTALLED; then echo cabal new-build -w ${HC} --disable-tests --disable-benchmarks $(${HCPKG} list --global --simple-output --names-only | sed 's/\([a-zA-Z0-9-]\{1,\}\) */--constraint="\1 installed" /g') all | sh; else echo "Not building with installed constraints"; fi
+  - if $INSTALLED; then echo cabal new-build -w ${HC} ${CABALFLAGS} --disable-tests --disable-benchmarks $(${HCPKG} list --global --simple-output --names-only | sed 's/\([a-zA-Z0-9-]\{1,\}\) */--constraint="\1 installed" /g') all | sh; else echo "Not building with installed constraints"; fi
 
   # build & run tests, build benchmarks
   - cabal new-build -w ${HC} ${CABALFLAGS} ${TEST} ${BENCH} -j2 all
-  - if [ "x$TEST" = "x--enable-tests" ]; then cabal new-test -w ${HC} ${TEST} ${BENCH} all; fi
+  - if [ "x$TEST" = "x--enable-tests" ]; then cabal new-test -w ${HC} ${CABALFLAGS} ${TEST} ${BENCH} all; fi
 
   # cabal check
   #   Commented out due to https://github.com/haskell/cabal/issues/4551
@@ -134,7 +134,7 @@ script:
 
   # haddock
   - rm -rf ./dist-newstyle
-  - if $HADDOCK; then cabal new-haddock -w ${HC} ${TEST} ${BENCH} all; else echo "Skipping haddock generation";fi
+  - if $HADDOCK; then cabal new-haddock -w ${HC} ${CABALFLAGS} ${TEST} ${BENCH} all; else echo "Skipping haddock generation";fi
 
 # REGENDATA ["-o",".travis.yml","cabal.project"]
 # EOF
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,7 +74,6 @@ install:
   - HADDOCK=${HADDOCK-true}
   - INSTALLED=${INSTALLED-true}
   - GHCHEAD=${GHCHEAD-false}
-  - CABALFLAGS="--flags=\"quickcheck-classes -aeson -semigroupoids\""
   - travis_retry cabal update -v
   - "sed -i.bak 's/^jobs:/-- jobs:/' ${HOME}/.cabal/config"
   - rm -fv cabal.project cabal.project.local
@@ -118,14 +117,14 @@ script:
   - "printf 'packages: primitive-*/*.cabal primitive-tests-*/*.cabal\\n' > cabal.project"
   - cat cabal.project
   # this builds all libraries and executables (without tests/benchmarks)
-  - cabal new-build -w ${HC} ${CABALFLAGS} --disable-tests --disable-benchmarks all
+  - cabal new-build -w ${HC} --disable-tests --disable-benchmarks all
 
   # Build with installed constraints for packages in global-db
-  - if $INSTALLED; then echo cabal new-build -w ${HC} ${CABALFLAGS} --disable-tests --disable-benchmarks $(${HCPKG} list --global --simple-output --names-only | sed 's/\([a-zA-Z0-9-]\{1,\}\) */--constraint="\1 installed" /g') all | sh; else echo "Not building with installed constraints"; fi
+  - if $INSTALLED; then echo cabal new-build -w ${HC} --disable-tests --disable-benchmarks $(${HCPKG} list --global --simple-output --names-only | sed 's/\([a-zA-Z0-9-]\{1,\}\) */--constraint="\1 installed" /g') all | sh; else echo "Not building with installed constraints"; fi
 
   # build & run tests, build benchmarks
-  - cabal new-build -w ${HC} ${CABALFLAGS} ${TEST} ${BENCH} -j2 all
-  - if [ "x$TEST" = "x--enable-tests" ]; then cabal new-test -w ${HC} ${CABALFLAGS} ${TEST} ${BENCH} all; fi
+  - cabal new-build -w ${HC} ${TEST} ${BENCH} -j2 all
+  - if [ "x$TEST" = "x--enable-tests" ]; then cabal new-test -w ${HC} ${TEST} ${BENCH} all; fi
 
   # cabal check
   #   Commented out due to https://github.com/haskell/cabal/issues/4551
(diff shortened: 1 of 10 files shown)


Data/Primitive/Array.hs
```diff
@@ -341,6 +343,11 @@ instance Eq a => Eq (Array a) where
    where loop i | i < 0     = True
                 | otherwise = indexArray a1 i == indexArray a2 i && loop (i-1)
 
+instance Eq1 Array where
+  liftEq p a1 a2 = sizeofArray a1 == sizeofArray a2 && loop (sizeofArray a1 - 1)
+   where loop i | i < 0     = True
+                | otherwise = p (indexArray a1 i) (indexArray a2 i) && loop (i-1)
+
```
treeowl: I just merged my eagerness patch, so you should use `indexArray##` here. Also, if we're adding `Eq1` we should surely also add `Ord1`.

andrewthad:
@treeowl  I probably won’t be able to get back to this PR tomorrow.  If you want, go ahead and take over it.  I am pretty confident that we are in agreement about the improvements that still need to be made to it.

treeowl:
I'm not sure when I'll be able to get to it either! Things are going to be
pretty crazy for the next couple weeks, TBH.

andrewthad:
It turns out that the `IsList` instance for `ByteArray` was wrong since `foldrByteArray` was wrong.

andrewthad:
@treeowl For some reason, github won't show me which code your comment refers to, so I'll try to explain more generally what's going on. On GHC 7.4.2, there is no `IsList` typeclass, so the `Data.Primitive.SmallArray` module doesn't end up exporting anything that let's us build a `SmallArray` from a list. Consequently, I have to copy their definitions into the test suite.

andrewthad:
@RyanGlScott This appears to be passing in travis now. There's a bogus failure for GHC 7.8, but everything else is good. Please review when you get a chance. In the process of adding the test suite, it was discovered that `<*>` was incorrect for both `Array` and `SmallArray` and that the `IsList` instance for `ByteArray` was incorrect. These have been fixed so that the test suite would pass.

primitive.cabal
```diff
@@ -52,7 +52,7 @@ Library
 
   Build-Depends: base >= 4.5 && < 4.12
                , ghc-prim >= 0.2 && < 0.6
-               , transformers >= 0.2 && < 0.6
+               , transformers >= 0.4 && < 0.6
```
RyanGlScott: Why are you bumping the upper version bounds for `transformers`? (Note that that is what is causing the GHC 7.8 build failure in Travis.)

test/main.hs
```diff
@@ -69,3 +119,46 @@ mkByteArray xs = runST $ do
...
+iforM_ xs0 f = go 0 xs0 where
+  go !_ [] = return ()
+  go !ix (x : xs) = f ix x >> go (ix + 1) xs
+
+infixl 1 ?
+(?) :: (a -> b -> c) -> (b -> a -> c)
+(?) = flip
+{-# INLINE (?) #-}
+
+smallArrayFromListN :: Int -> [a] -> SmallArray a
```
RyanGlScott: If this is literally copy-pasted from the main library, it might make sense to just export it from `Data.Primitive.SmallArray` proper.

primitive.cabal
treeowl: If we can keep the bounds down, maybe we can just deprecate the last couple versions altogether. The correctness and safety bugs are really piling up.

Data/Primitive/ByteArray.hs
```diff
@@ -181,12 +180,16 @@ foldrByteArray f z arr = go 0
...
-fromListN n xs = runST $ do
-    marr <- newByteArray (n * sizeOf (head xs))
-    zipWithM_ (writeByteArray marr) [0..n] xs
+fromListN n ys = runST $ do
+    marr <- newByteArray (n * sizeOf (head ys))
+    let go !_ [] = return ()
+        go !ix (x : xs) = do
+          writeByteArray marr ix x
+          go (ix + 1) xs
+    go 0 ys
```
treeowl: While you're replacing the `IsList` instance, I think you should make it safe by checking against the given size on each iteration. I'm adding that check to the polymorphic arrays.

treeowl:
Oh, I see... You want newer transformers because the lifted classes changed... Hrmm. I think it might be better just not to run those tests with sufficiently elderly GHC versions. Or something. Gross. Can we avoid `Eq1`, etc., in the test suite by using `FlexibleContexts`?

treeowl:
@RyanGlScott Who is going to decide what happens with the last two releases? Having bug-ridden instances in non-deprecated releases seems rather bad.

RyanGlScott:
I don't have Hackage permissions for `primitive`, so there's nothing I can do about that.

treeowl:
Ah, I guess @dolio will have to chime in at some point.

andrewthad:
I'll need to think a little more about how to allow `transformers-0.2` and `transformers-0.3`. I don't particularly want to change `quickcheck-classes` to use `FlexibleContexts` since that makes the type signatures less clear. I'll keep thinking about this and see if I can come up with a different solution.

Data/Primitive/ByteArray.hs
andrewthad: I’ll do that.  It’s a pretty inexpensive check, and this function  should never be used in performance-critical situations, and segfaults  are terrible.

andrewthad:
So, what I think that I’m going to do is make quickcheck-classes  able to build with even older versions of transformers.  When built with these old versions of transformers (and a sufficiently old version of base) , it will not provide property tests for any of the higher kinded type classes. In primitive, I’ll just CPP our the functor, applicative, etc. tests when the dependencies are old.  Does this sound like an acceptable approach?  Practically speaking, I believe that the effect it will have will be that GHC 7.8.4  does not get  tests for higher kinded type classes run (since that appears to be the only one where transformers < 0.4 is used).

andrewthad:
Dang. New versions of quickcheck all require transformers >= 0.3, so that approach isn't feasible.

RyanGlScott:
I think 0.3 would be an acceptable lower bound. We only need 0.3 since it's shipped with 7.8 itself. As far as I can tell, we could do away with 0.2 if need be (warning: untested hypothesis).

andrewthad:
Right after pushing that last commit, I noticed @treeowl's work on https://github.com/haskell/primitive/pull/100. Sorry for the duplication. On the upside, it looks like we came up with the exact same implementation.

treeowl:
@andrewthad I'm actually working on another version that tries to fuse with list producers under certain circumstances. Is there a reason for the reimplementation to be in this PR? If not, I'd prefer to keep it separate.

andrewthad:
Ah, cool. Yeah, definitely no reason for that to be in this PR.

andrewthad:
Aw yeah, look at that green checkmark next to the travis build. This now builds with `transformers-0.3.0.0` again.

andrewthad:
Actually, we could bump the transformers lower bound back down to `0.2.0.0`. The `primitive` library should build just fine (there's no CPP that distinguishes `transformers` 0.2 and 0.3). It's only the test suite that cannot be built in this case. Somewhat fortunately, none of the travis tests attempt to build against transformers-0.2.0.0, so it would still pass.

test/primitive-tests.cabal
```diff
@@ -31,6 +31,12 @@ test-suite test
   build-depends: base >= 4.5 && < 4.12
                , ghc-prim
                , primitive
+               , QuickCheck
+               , tasty
+               , tasty-quickcheck
+               , tagged
+               , transformers >= 0.3
+               , quickcheck-classes == 0.4.2
```
RyanGlScott: To make this more robust in the future, let's do `>= 0.4.2` instead.

Data/Primitive/SmallArray.hs
```diff
@@ -797,3 +848,25 @@ instance (Typeable s, Typeable a) => Data (SmallMutableArray s a) where
   gunfold _ _ = die "gunfold" "SmallMutableArray"
   dataTypeOf _ = mkNoRepType "Data.Primitive.SmallArray.SmallMutableArray"
 #endif
+
+smallArrayFromListN :: Int -> [a] -> SmallArray a
```
RyanGlScott: Since we're exporting this, it would be good to add at least some documentation for what this is doing.

Data/Primitive/SmallArray.hs
```diff
@@ -797,3 +848,25 @@ instance (Typeable s, Typeable a) => Data (SmallMutableArray s a) where
...
+          writeSmallArray sma ix x
+          go (ix+1) xs
+        else die "smallArrayFromListN" "list length greater than specified size"
+  go 0 l
+  unsafeFreezeSmallArray sma
+#else
+smallArrayFromListN n l = SmallArray (Array.fromListN n l)
+#endif
+
+smallArrayFromList :: [a] -> SmallArray a
```
RyanGlScott: Similarly, this is also exported, so give some Haddocks for it.

Data/Primitive/Array.hs
```diff
@@ -63,6 +63,10 @@ import Data.Functor.Identity
 
 import Text.ParserCombinators.ReadP
 
+#if MIN_VERSION_base(4,9,0) || MIN_VERSION_transformers(0,4,0)
+import Data.Functor.Classes (Eq1(..),Ord1(..),Show1(..))
```
RyanGlScott: This leaves out `Read1`... was this intentional?

changelog.md
```diff
@@ -1,3 +1,18 @@
+## Changes in version 0.6.4.0
+
+ * Add `Eq1`, `Ord1`, and `Show1` instances for `Array` and `SmallArray`.
+
+ * Drop support for `transformers-0.2.x.y`.
```
RyanGlScott: This can be dropped (I hope).

Data/Primitive/SmallArray.hs
andrewthad: Will do.

Data/Primitive/SmallArray.hs
andrewthad: Will do.

test/primitive-tests.cabal
andrewthad: I'm a little worried about accidentally break `primitive`'s test suite with a future release of `quickcheck-classes`, but I'll go ahead and do this.

RyanGlScott: If that is a concern, it might be worthwhile to add an upper version bound of `< 0.5`.

changelog.md
andrewthad: I'll drop this.

Data/Primitive/Array.hs
andrewthad: I'll add this.

andrewthad:
Just pushed more changes addressing @RyanGlScott's most recent comments. I've decided to leave the quickcheck-classes upper bound off for the time being.

changelog.md
```diff
@@ -1,3 +1,16 @@
+## Changes in version 0.6.4.0
+
+ * Add `Eq1`, `Ord1`, and `Show1` instances for `Array` and `SmallArray`.
```
RyanGlScott: Don't forget to mention `Read1` here!

andrewthad:
Just pushed a commit (and a new version of `quickcheck-classes`) where all the `Traversable` laws are tested. Also, I'm testing `foldl1` and `foldr1` now. As you suspected, something was still wrong with `foldl1` (it was doing a right-associated fold instead of left-associated), and I've fixed that.

andrewthad:
Build is passing.

Data/Primitive/SmallArray.hs
```diff
@@ -111,6 +116,11 @@ newtype SmallArray a = SmallArray (Array a) deriving
   , MonadFix
   , Monoid
   , Typeable
+#if MIN_VERSION_base(4,9,0) || MIN_VERSION_transformers(0,4,0)
+  , Eq1
+  , Ord1
```
RyanGlScott: Perhaps we should derive `Read1` here as well?

andrewthad: Your thoroughness knows no bounds.

RyanGlScott:
Excellent work, @andrewthad.

At this point, I really only have one additional comment, which is about dependencies. Due to the hacky nature of `primitive`'s test suite, we cannot cache the builds of any library which depends on `primitive`. Thus, with the addition of `quickcheck-instances`, we have to build several additional libraries on _every_ run of Travis, including `semigroupoids` and `aeson`, which are somewhat bulky.

Perhaps this is an acceptable cost—I'm not sure. But it is worth thinking about. What are your thoughts?

andrewthad:
@RyanGlScott It looks like the deps on `semigroupoids` and `aeson` only come from `quickcheck-classes`. I can add a cabal flag for turning those off. Do you know how to get travis to pass a flag to a dependency? (I don't know how to use travis for cabal builds very well but I assume this is possible).

RyanGlScott:
@andrewthad, that might be a worthwhile pursuit.

As far as how to change Travis, I think it would be a matter of passing `--flags='quickcheck-classes -aeson -semigroupoids'` to each invocation of `cabal new-*` in the script. The cleanest way to do this would be to declare:

```
- CABALFLAGS="--flags='quickcheck-classes -aeson -semigroupoids'"
```

Somewhere around [here](https://github.com/haskell/primitive/blob/ea0734a279481f0a04d59f020a41f9f4af872d47/.travis.yml#L76) and pass `${CABALFLAGS}` to the right places.

.travis.yml
```diff
@@ -123,7 +124,7 @@ script:
   - if $INSTALLED; then echo cabal new-build -w ${HC} --disable-tests --disable-benchmarks $(${HCPKG} list --global --simple-output --names-only | sed 's/\([a-zA-Z0-9-]\{1,\}\) */--constraint="\1 installed" /g') all | sh; else echo "Not building with installed constraints"; fi
 
   # build & run tests, build benchmarks
-  - cabal new-build -w ${HC} ${TEST} ${BENCH} -j2 all
+  - cabal new-build -w ${HC} ${CABALFLAGS} ${TEST} ${BENCH} -j2 all
```
RyanGlScott: You'll need to pass `${CABALFLAGS}` to more than just here. You'll also need it on lines 121, 124, 128, and 137, which call `new-build`, `new-test`, and `new-haddock`. (Otherwise, when those commands are run, it will first calculate a new set of dependencies, since you passed a different set of flags!)

andrewthad: Ah, thanks. I had forgotten that `new-test` even existed, so I didn't think to search for that. Also, it's a little strange to me that this to the earlier `new-build` commands in the file (the ones that don't build suite) would have any affect at all since `primitive` doesn't actually depend on `quickcheck-classes`.

RyanGlScott: To be clear, it's not necessarily guaranteed that not passing `${CABALFLAGS}` would make a difference. But in my experience, things tend to work best when you use a consistent set of build flags everywhere—you don't want to give the recompilation checker any excuse for believing that some dependency needs to be rebuilt.

.travis.yml
```diff
@@ -74,6 +74,7 @@ install:
   - HADDOCK=${HADDOCK-true}
   - INSTALLED=${INSTALLED-true}
   - GHCHEAD=${GHCHEAD-false}
+  - CABALFLAGS="--flags='quickcheck-classes -aeson -semigroupoids'"
```
RyanGlScott: I think I goofed—it turns out `cabal-install` doesn't like it when you surround this with double-quotes. I think it should work if you remove them.

andrewthad:
I think I actually need to use `constraint` instead of `flags`. Trying this again.

RyanGlScott:
No, `flags` should be correct. But I realized another issue—the quotes around the arguments to `flags` aren't being preserved. This should _actually_ do the trick:

```
CABALFLAGS="--flags=\"quickcheck-classes -aeson -semigroupoids\""
```

I actually tested this one, so it better work :)

andrewthad:
No dice.

RyanGlScott:
Ugh. When I said "I actually tested this one", I meant I ran `echo ${CABALFLAGS}`, and it produced the output I expected. But evidently, this isn't what is actually passed to other commands... god I hate `bash`.

I'll experiment more with this locally.

RyanGlScott:
On second thought, we might be making this way more complicated than need be. I neglected to notice that you actually amended `cabal.project` with the right flags in the first place! In light of that, I think this `CABALFLAGS` hack is wholly unnecessary. Can you try removing it and seeing if that does the right thing?

Sorry for the agonizing delay, @andrewthad—this stuff is always trickier than one would hope.

RyanGlScott:
Oh, right. I forgot that `haskell-ci` generates its _own_ `cabal.project` on the fly, which is why putting those flags in the local `cabal.project` won't show up. I know how to fix this, but this is something that can be done after merging.

In any case, this is building, and we should get this landed. Thanks for all the hard work, @andrewthad!

andrewthad:
Thanks!

treeowl:
No, thank *you*. Testing is some of the least visible and most important maintenance work there is.

PR #100, opened 2018-03-21
Make Array and SmallArray IsList instances safe

treeowl:
* Don't overrun the array bounds if someone calls
  `fromListN` with an incorrect size argument.

* Implement list fusion for `fromListN`. We can probably
  improve this, but it's a decent draft.

### DIFF
--- a/Data/Primitive/Array.hs
+++ b/Data/Primitive/Array.hs
@@ -43,7 +43,7 @@ import Data.Primitive.Internal.Compat ( isTrue#, mkNoRepType )
 import Control.Monad.ST(ST,runST)
 
 import Control.Applicative
-import Control.Monad (MonadPlus(..), when)
+import Control.Monad (MonadPlus(..), when, unless)
 import Control.Monad.Fix
 #if MIN_VERSION_base(4,4,0)
 import Control.Monad.Zip
@@ -59,6 +59,9 @@ import Data.Semigroup
 #endif
 #if MIN_VERSION_base(4,8,0)
 import Data.Functor.Identity
+#endif
+#if !(MIN_VERSION_base(4,8,0))
+import Prelude hiding (foldr)
 #endif
 
 import Text.ParserCombinators.ReadP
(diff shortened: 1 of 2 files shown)


cartazio:
Where’s the RULES for  fusion?

treeowl:
@cartazio This version just inlines to something that uses `foldr` and should fuse. *However*, I decided we shouldn't do it like this. The trouble is that the typical use of `fromListN` is for `-XOverloadedLists`, and list literals won't fuse anyway. So it would be better not to end up with the conversion code uselessly inlined in that case. I'm working on a version that breaks things down into smaller pieces and uses `RULES`; it's just not ready yet.

treeowl:
@cartazio the new version I haven't pushed yet is the way this sort of thing is generally done in `base`. But I'm going to need to refine it a little to avoid over-inlining with `augment` (e.g., `fromListN 5 ([1..3] ++ [7,9])`).

treeowl:
Hmm.... I'm struggling to get the write-back rule to work with `augment`. It looks like GHC probably isn't beta reducing something I want it to, despite an `INLINE CONLIKE` pragma.

treeowl:
I'm not sure I'll be able to get write-back rules to work. Something is blocking that up, and I think it's likely GHC's desire to avoid duplicating work in beta reduction. So we can use this version if we like, or do the kind of gross (but very well-behaved) thing of adding rules to fuse specifically with `build` and `augment`.

For the record, I should mention why safety matters to me here, in a generally unsafe library. As long as all class instances are safe, `Trustworthy` modules can re-export the array types (and constructors). But when instances are unsafe, they can only export wrapped versions.

cartazio:
lets do the simple correct ones for now, and then factor out making it faster / better optimized later

cartazio:
should the fromList/IsList stuff be INLINE if we're not gonna depend on fusion?

treeowl:
Why not try for performance? This is one of the most convenient ways to
build an array! If we're not doing fusion, I don't think it should be
INLINE, no. GHC will probably inline it anyway, but whatever.

cartazio:
perhaps your right, :)

treeowl:
They're safe now; I'll try to open another PR for fusion.

PR #104, opened 2018-03-23
Make array creation friendlier to worker-wrapper

### LINKED ISSUE #102: Rethink the empty array optimization
Currently,

```haskell
createArray 0 _ _ = emptyArray
createArray n x f = runST $ do
  ma <- newArray n x
  f ma
  unsafeFreezeArray ma
```

This is very straightforward, but it doesn't play particularly well with worker-wrapper. In particular, whenever the array is non-empty, it will allocate both an array and an `Array` constructor to hold it. Since the empty array is the *unusual* case, I think we might be better off optimizing the non-empty case more aggressively at its expense:

```haskell
createArray
  :: Int
  -> a
  -> (forall s. MutableArray s a -> ST s ())
  -> Array a
createArray n x f = Array (createArray# n x f)

createArray#
  :: Int
  -> a
  -> (forall s. MutableArray s a -> ST s ())
  -> Array# a
createArray# 0 _ _ = emptyArray# (# #)
createArray# (I# n) x f = case runRW# $ \s ->
   case newArray# n x s of { (# s', mary# #) ->
   case unST (f (MutableArray mary#)) s' of { (# s'', _ #) ->
   unsafeFreezeArray# mary# s'' }} of (# _, ary# #) -> ary#

unST :: ST s a -> State# s -> (# State# s, a #)
unST (ST f) = f

emptyArray# :: (# #) -> Array# a
emptyArray# _ = case emptyArray of Array ar -> ar
{-# NOINLINE emptyArray# #-}
```

treeowl:
Closes #102

### DIFF
--- a/Data/Primitive/Array.hs
+++ b/Data/Primitive/Array.hs
@@ -17,7 +17,7 @@ module Data.Primitive.Array (
   Array(..), MutableArray(..),
 
   newArray, readArray, writeArray, indexArray, indexArrayM,
-  freezeArray, thawArray,
+  freezeArray, thawArray, runArray,
   unsafeFreezeArray, unsafeThawArray, sameMutableArray,
   copyArray, copyMutableArray,
   cloneArray, cloneMutableArray,
@@ -54,11 +54,17 @@ import Data.Traversable (Traversable(..))
 import Data.Monoid
 #endif
 #if MIN_VERSION_base(4,9,0)
+import qualified GHC.ST as GHCST
 import qualified Data.Foldable as F
 import Data.Semigroup
 #endif
 #if MIN_VERSION_base(4,8,0)
 import Data.Functor.Identity
+#endif
+#if MIN_VERSION_base(4,10,0)
+import GHC.Exts (runRW#)
+#elif MIN_VERSION_base(4,9,0)
+import GHC.Base (runRW#)
 #endif
 
 import Text.ParserCombinators.ReadP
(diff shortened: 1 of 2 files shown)


treeowl:
The Core looks rather better, IMO (see, e.g., what this does to `fmap`). I don't have benchmarks to prove it's good, but I think we should merge this anyway.

treeowl:
Another option is to write `runArray` in this manner (which is smaller and simpler), and use that to implement `createArray`. The inlining decisions end up a little different; I'm not sure which would really be better. But we likely should export `runArray` anyway.

RyanGlScott:
Forgive me for being dense, but I'm still a bit unclear (from the comments, at least) how this is improving the interaction w.r.t. worker-wrapper. Can you expand on the comments a bit to clarify?

Also, if I understand correctly, this will make creating empty arrays slightly slower, yes? If so, is there an alternative way to create an empty array that is faster? We should point in out in the Haddocks if so.

cartazio:
i guess another way of stating Ryan's question is:

what is a small example of what worker wrapper will facilitate here?

An important detail to keep in mind with Primitive is that its *foremost* a toolkit for mutable data structures, and fusion-y optimizations tend to be very low ROI in those cases.  We should provide all the *right* primitives for supporting easy optimization in client libraries (like vector), but that doesn't mean we should push fusion into this layer?

I'm open to being educated / convinced about this, but could you help Ryan and Myself understand the goal / code improvements this will facilitate and some toy example where the change is high impact/measurable?

treeowl:
The fact that immutable arrays aren't the prime directive doesn't mean they should get short shrift! I'm not really doing anything fancy here. Suppose I write `f (fmap g ary)`, where `f` is strict (it's not just sticking the array in an array or something silly like that). We don't generally want `fmap` to actually produce an `Array` here. We want to produce an `Array#` and pass that straight to the worker for `f`. The problem with the current implementation is that it applies the `Array` constructor inside an `ST` action. `runRW#` doesn't inline till Core Prep, when it's too late to unbox anything. By moving the `Array` constructor application outside `runRW#`, we let GHC erase it.

treeowl:
@RyanGlScott, yes, this will make creating empty arrays with `createArray` a bit slower. Instead of pointing to the same `Array`, separate invocations will point to different `Array`s that all point to the same `Array#`. Since `createArray` is internal, I don't think there's anything to be gained from an extra empty export.

RyanGlScott:
> Since `createArray` is internal

Oops, I missed that part, so that makes that concern pretty much moot :)

Issue #58, opened 2017-06-13
ByteArray with phantom type variable


andrewthad:
One difficult of using `ByteArray` is that some of the functions interpret index and length arguments to mean "number of bytes", and others interpret it to mean "number of elements":

- bytes: newByteArray, copyByteArray, sizeofByteArray, etc.
- elements: writeByteArray, fillByteArray, etc.

This means that users of `ByteArray` and `MutableByteArray` have to keep track of whether or not they they need to multiply an index by the size of the elements.

I've been working on a project where I wrote [`PrimArray` and `MutablePrimArray`](https://github.com/andrewthad/btree/blob/b4e9d9c1fcfb1af0e12ad7d52be3cadee23e3053/src/Data/Primitive/PrimArray.hs), which are the same thing as `ByteArray` and `MutableByteArray` except that they also take a phantom type variable. This means that all functions operating on them can consistently talk about indexing and length in terms of elements, not bytes. It also prevents some mistakes that can happen when a user accidentally writes the wrong type of element to a `ByteArray`.

The module is not yet fully fleshed out, but once it is, I was going to move it into its own library and release it. However, I would prefer to add it to `primitive` if the library authors feel like it makes sense to have here. Just let me know.

vincenthz:
@andrewthad there's a similar thing to what you describe here: https://github.com/haskell-foundation/foundation/blob/master/Foundation/Primitive/Block.hs

dolio:
I suppose my question would be: what the difference is between these and `Data.Vector.Primitive(.Mutable)`?

winterland1989:
I made an implementation with the respect to `Data.Array` interface during my experiment: https://github.com/winterland1989/stdio/blob/master/Data/Primitive/PrimArray.hs

If someone want this, i can offer a PR.

andrewthad:
I've also got a package that provides this: http://hackage.haskell.org/package/prim-array. I think @winterland1989 has more things implemented though.

@dolio The difference is that this doesn't track a start index and length. Practically, this means that we can do things like store a `PrimArray` in an `ArrayArray`, which is not possible with anything from `vector`.

andrewthad:
@dolio Additionally, `PrimArray` outperforms `Data.Vector.Primitive` at very specific tasks since you don't have add the start position every time you index into the array. Like I said, only for very specific tasks does this matter.

cartazio:
With O2 Or with static size?

bgamari:
[...]

PR #106, opened 2018-03-23
Implement PrimArray

andrewthad:
This was original requested in https://github.com/haskell/primitive/issues/58 and was received positively. Indeed, it appears that at three different authors have built nearly the same thing in the wild. This feature was implemented as a part of the larger and more ambitious https://github.com/haskell/primitive/pull/64, which eventually stalled. As @treeowl [commented](https://github.com/haskell/primitive/pull/64):

> This pull request is very large. Lots of new functionality. I suspect it might be better to break it up.

This is an attempt at precisely that. All that is implemented here is `PrimArray`. Discussion of design choices:

- In this PR, `PrimArray` is implemented as a data type on top of `ByteArray#`, not as a newtype on top of `ByteArray`. This makes some of the `Monoid` and `Semigroup` functions allocate a little more than they might otherwise (using `coerce` to sidestep the allocations). I don't think this matters though, since using `mconcat` on a `ByteArray`/`PrimArray` is an indication that you already built your array suboptimally.
- The type variable associated with a `PrimArray` should probably have a nominal role. (so maybe the `coerce` trick for newtypes couldn't actually work)
- There are functions for copying between `PrimArray` and `Ptr`. We have the analogy: `PrimArray` is to `Ptr` as `ByteArray` is to `Addr`.
- The `Eq` instance is intentionally the slow element-by-element comparison. It doesn't use `memcmp` under the hood because someone may want to define a type whose `Eq` instance evaluate values with different in-memory representations as equivalent. (ie- modular arithmetic)
- We get to have a much nicer `FromList` instance than `ByteArray`.

### DIFF
--- a/Data/Primitive/PrimArray.hs
+++ b/Data/Primitive/PrimArray.hs
@@ -70,6 +70,7 @@ module Data.Primitive.PrimArray
 
 import GHC.Prim
 import GHC.Base ( Int(..) )
+import GHC.Exts (build)
 import GHC.Ptr
 import Data.Primitive.Internal.Compat (isTrue#)
 import Data.Primitive
@@ -150,8 +151,9 @@ primArrayFromListN len vs = runST run where
     go vs 0
     unsafeFreezePrimArray arr
 
+{-# INLINE primArrayToList #-}
 primArrayToList :: forall a. Prim a => PrimArray a -> [a]
-primArrayToList = foldrPrimArray (:) []
+primArrayToList xs = build (\c n -> foldrPrimArray c n xs)
 
 primArrayToByteArray :: PrimArray a -> PB.ByteArray
 primArrayToByteArray (PrimArray x) = PB.ByteArray x
--- a/Data/Primitive/PrimArray.hs
+++ b/Data/Primitive/PrimArray.hs
@@ -15,6 +15,9 @@ module Data.Primitive.PrimArray
     -- * Allocation
   , newPrimArray
   , resizeMutablePrimArray
+#if __GLASGOW_HASKELL__ >= 710
+  , shrinkMutablePrimArray
+#endif
     -- * Element Access
   , readPrimArray
   , writePrimArray
@@ -193,7 +196,7 @@ newPrimArray (I# n#)
         (# s'#, arr# #) -> (# s'#, MutablePrimArray arr# #)
     )
 
--- | Resize a mutable primitive array. The new size is given in bytes.
+-- | Resize a mutable primitive array. The new size is given in elements.
 --
 -- This will either resize the array in-place or, if not possible, allocate the
 -- contents into a new, unpinned array and copy the original array\'s contents.
@@ -217,6 +220,21 @@ resizeMutablePrimArray arr n
   = do arr' <- newPrimArray n
        copyMutablePrimArray arr 0 arr' 0 (min (sizeofMutablePrimArray arr) n)
        return arr'
+#endif
+
+-- Although it is possible to shim resizeMutableByteArray for old GHCs, this
+-- is not the case with shrinkMutablePrimArray.
+#if __GLASGOW_HASKELL__ >= 710
+-- | Shrink a mutable primitive array. The new size is given in elements.
+-- It must be smaller than the old size. The array will be resized in place.
+-- This function is only available when compiling with GHC 7.10 or newer.
+shrinkMutablePrimArray :: forall m a. (PrimMonad m, Prim a)
+  => MutablePrimArray (PrimState m) a
+  -> Int -- ^ new size
+  -> m ()
+{-# INLINE shrinkMutablePrimArray #-}
+shrinkMutablePrimArray (MutablePrimArray arr#) (I# n#)
+  = primitive_ (shrinkMutableByteArray# arr# (n# *# sizeOf# (undefined :: a)))
 #endif
 
 readPrimArray :: (Prim a, PrimMonad m) => MutablePrimArray (PrimState m) a -> Int -> m a
@@ -340,7 +358,9 @@ getSizeofMutablePrimArray arr
   = return (sizeofMutablePrimArray arr)
 #endif
 
--- | Size of the mutable primitive array in elements.
+-- | Size of the mutable primitive array in elements. This function shall not
+--   be used on primitive arrays that are an argument to or a result of
+--   'resizeMutablePrimArray' or 'shrinkMutablePrimArray'.
 sizeofMutablePrimArray :: forall s a. Prim a => MutablePrimArray s a -> Int
 {-# INLINE sizeofMutablePrimArray #-}
 sizeofMutablePrimArray (MutablePrimArray arr#) =
--- a/Data/Primitive/PrimArray.hs
+++ b/Data/Primitive/PrimArray.hs
@@ -61,11 +61,14 @@ module Data.Primitive.PrimArray
   , generatePrimArrayA
   , replicatePrimArrayA
   , filterPrimArrayA
+  , mapMaybePrimArrayA
     -- ** Strict Primitive Monadic
   , traversePrimArrayP
   , itraversePrimArrayP
   , generatePrimArrayP
   , replicatePrimArrayP
+  , filterPrimArrayP
+  , mapMaybePrimArrayP
   ) where
 
 import GHC.Prim
@@ -524,6 +527,50 @@ traversePrimArrayP f arr = do
   go 0
   unsafeFreezePrimArray marr
 
+{-# INLINE filterPrimArrayP #-}
+filterPrimArrayP :: (PrimMonad m, Prim a)
+  => (a -> m Bool)
+  -> PrimArray a
+  -> m (PrimArray a)
+filterPrimArrayP f arr = do
+  let !sz = sizeofPrimArray arr
+  marr <- newPrimArray sz
+  let go !ixSrc !ixDst = if ixSrc < sz
+        then do
+          let a = indexPrimArray arr ixSrc
+          b <- f a
+          if b
+            then do
+              writePrimArray marr ixDst a
+              go (ixSrc + 1) (ixDst + 1)
+            else go (ixSrc + 1) ixDst
+        else return ixDst
+  lenDst <- go 0 0
+  marr' <- resizeMutablePrimArray marr lenDst
+  unsafeFreezePrimArray marr'
+
+{-# INLINE mapMaybePrimArrayP #-}
+mapMaybePrimArrayP :: (PrimMonad m, Prim a, Prim b)
+  => (a -> m (Maybe b))
+  -> PrimArray a
+  -> m (PrimArray b)
+mapMaybePrimArrayP f arr = do
+  let !sz = sizeofPrimArray arr
+  marr <- newPrimArray sz
+  let go !ixSrc !ixDst = if ixSrc < sz
+        then do
+          let a = indexPrimArray arr ixSrc
+          mb <- f a
+          case mb of
+            Just b -> do
+              writePrimArray marr ixDst b
+              go (ixSrc + 1) (ixDst + 1)
+            Nothing -> go (ixSrc + 1) ixDst
+        else return ixDst
+  lenDst <- go 0 0
+  marr' <- resizeMutablePrimArray marr lenDst
+  unsafeFreezePrimArray marr'
+
 {-# INLINE generatePrimArrayP #-}
 generatePrimArrayP :: (PrimMonad m, Prim a)
   => Int
@@ -630,6 +677,28 @@ filterPrimArrayA f = \ !ary ->
             (\keep (IxSTA m) -> IxSTA $ \ixDst mary -> if keep
               then writePrimArray (MutablePrimArray mary) ixDst x >> m (ixDst + 1) mary
               else m ixDst mary
+            )
+            (f x)
+            (go (ixSrc + 1))
+  in if len == 0
+     then pure emptyPrimArray
+     else runIxSTA len <$> go 0
+
+mapMaybePrimArrayA ::
+     (Applicative f, Prim a, Prim b)
+  => (a -> f (Maybe b)) -- ^ mapping function
+  -> PrimArray a -- ^ primitive array
+  -> f (PrimArray b)
+mapMaybePrimArrayA f = \ !ary ->
+  let
+    !len = sizeofPrimArray ary
+    go !ixSrc
+      | ixSrc == len = pure $ IxSTA $ \ixDst _ -> return ixDst
+      | otherwise = let x = indexPrimArray ary ixSrc in
+          liftA2
+            (\mb (IxSTA m) -> IxSTA $ \ixDst mary -> case mb of
+              Just b -> writePrimArray (MutablePrimArray mary) ixDst b >> m (ixDst + 1) mary
+              Nothing -> m ixDst mary
             )
             (f x)
             (go (ixSrc + 1))
--- a/Data/Primitive/PrimArray.hs
+++ b/Data/Primitive/PrimArray.hs
@@ -301,6 +301,9 @@ getSizeofMutablePrimArray (MutablePrimArray arr#)
         (# s'#, sz# #) -> (# s'#, I# (quotInt# sz# (sizeOf# (undefined :: a))) #)
     )
 #else
+-- On older GHCs, it is not possible to resize a byte array, so
+-- this provides behavior consistent with the implementation for
+-- newer GHCs.
 getSizeofMutablePrimArray arr
   = return (sizeofMutablePrimArray arr)
 #endif
--- a/Data/Primitive/PrimArray.hs
+++ b/Data/Primitive/PrimArray.hs
@@ -74,7 +74,7 @@ import GHC.Exts (build)
 import GHC.Ptr
 import Data.Primitive.Internal.Compat (isTrue#)
 import Data.Primitive
-import Data.Monoid (Monoid(..))
+import Data.Monoid (Monoid(..),(<>))
 import Control.Applicative
 import Control.Monad.Primitive
 import Control.Monad.ST
--- a/Data/Primitive/PrimArray.hs
+++ b/Data/Primitive/PrimArray.hs
@@ -494,19 +494,19 @@ foldlPrimArrayM' f z0 arr = go 0 z0
 -- the performance. Consider the following short-circuiting traversal:
 --
 -- > incrPositiveA :: PrimArray Int -> Maybe (PrimArray Int)
--- > incrPositiveA xs = traversePrimArray (\x -> bool Nothing (Just x + 1) (x > 0)) xs
+-- > incrPositiveA xs = traversePrimArray (\x -> bool Nothing (Just (x + 1)) (x > 0)) xs
 --
 -- This can be rewritten using 'traversePrimArrayP'. To do this, we must
 -- change the traversal context to @MaybeT (ST s)@, which has a 'PrimMonad'
 -- instance:
 --
 -- > incrPositiveB :: PrimArray Int -> Maybe (PrimArray Int)
--- > incrPositiveB xs = runST $ traversePrimArrayP
+-- > incrPositiveB xs = runST $ runMaybeT $ traversePrimArrayP
 -- >   (\x -> bool (MaybeT (return Nothing)) (MaybeT (return (Just (x + 1)))) (x > 0))
 -- >   xs
 -- 
--- Benchmarks demonstrate that the second implementation outperforms
--- the first by a factor of X.
+-- Benchmarks demonstrate that the second implementation runs 150 times
+-- faster than the first. It also results in fewer allocations.
 {-# INLINE traversePrimArrayP #-}
 traversePrimArrayP :: (PrimMonad m, Prim a, Prim b)
   => (a -> m b)
--- a/Data/Primitive/PrimArray.hs
+++ b/Data/Primitive/PrimArray.hs
@@ -396,8 +396,14 @@ foldlPrimArrayM' f z0 arr = go 0 z0
           go (i + 1) acc2
       | otherwise = return acc1
 
--- | Monadic map over a 'PrimArray' where the monad allows
---   primitive actions.
+-- | Traverse a primitive array. The traversal forces the resulting values and
+-- writes them to the new primitive array as it performs the monadic effects.
+-- Consequently:
+--
+-- >>> traversePrimArrayP (\x -> print x $> bool x undefined (x == 2)) (fromList [1, 2, 3 :: Int])
+-- 1
+-- 2
+-- *** Exception: Prelude.undefined
 {-# INLINE traversePrimArrayP #-}
 traversePrimArrayP :: (PrimAffineMonad m, Prim a, Prim b)
   => (a -> m b)
@@ -445,6 +451,7 @@ generatePrimArrayUnsafe sz f = do
   go 0
   unsafeFreezePrimArray marr
 
+-- | Map over the elements of a primitive array.
 {-# INLINE mapPrimArray #-}
 mapPrimArray :: (Prim a, Prim b)
   => (a -> b)
@@ -462,6 +469,7 @@ mapPrimArray f arr = runST $ do
   go 0
   unsafeFreezePrimArray marr
 
+-- | Indexed map over the elements of a primitive array.
 {-# INLINE imapPrimArray #-}
 imapPrimArray :: (Prim a, Prim b)
   => (Int -> a -> b)
@@ -479,11 +487,23 @@ imapPrimArray f arr = runST $ do
   go 0
   unsafeFreezePrimArray marr
 
--- | Traverse a primitive array.
+-- | Traverse a primitive array. The traversal performs all of the applicative
+-- effects /before/ forcing the resulting values and writing them to the new
+-- primitive array. Consequently:
+--
+-- >>> traversePrimArray (\x -> print x $> bool x undefined (x == 2)) (fromList [1, 2, 3 :: Int])
+-- 1
+-- 2
+-- 3
+-- *** Exception: Prelude.undefined
+--
+-- The function 'traversePrimArrayP' always outperforms this function, but it
+-- requires a 'PrimAffineMonad' constraint, and it forces the values as
+-- it performs the effects.
 traversePrimArray ::
      (Applicative f, Prim a, Prim b)
-  => (a -> f b)
-  -> PrimArray a
+  => (a -> f b) -- ^ mapping function
+  -> PrimArray a -- ^ primitive array
   -> f (PrimArray b)
 traversePrimArray f = \ !ary ->
   let
@@ -541,6 +561,7 @@ itraversePrimArrayUnsafe f arr = do
   go 0
   unsafeFreezePrimArray marr
 
+-- | Generate a primitive array 
 {-# INLINE generatePrimArray #-}
 generatePrimArray :: Prim a
   => Int -- ^ length
--- a/Data/Primitive/PrimArray.hs
+++ b/Data/Primitive/PrimArray.hs
@@ -58,6 +58,7 @@ module Data.Primitive.PrimArray
   , itraversePrimArray
   , generatePrimArrayA
   , replicatePrimArrayA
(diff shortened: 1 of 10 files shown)


Data/Primitive/PrimArray.hs
```diff
@@ -0,0 +1,269 @@
...
+data PrimArray a = PrimArray ByteArray#
+
+-- | Mutable primitive arrays associated with a primitive state token
+data MutablePrimArray s a = MutablePrimArray (MutableByteArray# s)
+
+instance (Eq a, Prim a) => Eq (PrimArray a) where
+  a1 == a2 = sizeofPrimArray a1 == sizeofPrimArray a2 && loop (sizeofPrimArray a1 - 1)
+    where 
+    loop !i | i < 0 = True
+            | otherwise = indexPrimArray a1 i == indexPrimArray a2 i && loop (i-1)
```
treeowl: I think it's reasonable to require that the underlying `==` be an equivalence relation (here and for the other types). Reflexivity then allows us to do a quick pointer equality check with `sameWhateverArray`.

Data/Primitive/PrimArray.hs
```diff
@@ -0,0 +1,269 @@
...
+    arr <- newPrimArray len
+    let go :: [a] -> Int -> ST s ()
+        go !xs !ix = case xs of
+          [] -> return ()
+          a : as -> do
+            writePrimArray arr ix a
+            go as (ix + 1)
+    go vs 0
+    unsafeFreezePrimArray arr
+
```
treeowl: This needs a bounds check.

Data/Primitive/PrimArray.hs
```diff
@@ -0,0 +1,269 @@
...
+primArrayToList :: forall a. Prim a => PrimArray a -> [a]
+primArrayToList arr = go 0 where
+  !len = sizeofPrimArray arr
+  go :: Int -> [a]
+  go !ix = if ix < len
+    then indexPrimArray arr ix : go (ix + 1)
+    else []
```
treeowl: I suggest adding folds, and then copying the default definition of `toList` to implement this; that'll get you list fusion, and of course we want folds anyway.

Data/Primitive/PrimArray.hs
andrewthad: I think that's reasonable. Let me make sure I understand you correctly. You're saying that I would keep the `Prim` constraint and the currently implementation but would just a `sameByteArray` check that happens first to see if we can skip the element-by-element check. Is that correct?

treeowl: That's my suggestion, yes.

andrewthad:
@treeowl I've implemented your suggestions. Concerning folding functions, I'm holding off on introducing the whole slew of them because I want to get to the bottom of the question of naming. Do we want to continue down the route of suffixing every function name with the type of its argument? Do we want `foldlPrimArray`, `foldMapPrimArray`, etc., or would it be better to create a separate module that is intended to be imported qualified? (or I guess we could do both).

cartazio:
I like the mangled names.  Or at least let’s stick with it for now and
revisit it later

Unlike in vector primitive isn’t really a package where you will swap reps
that often.

treeowl:
I agree with @cartazio. I *dislike* the mangled names, but I think trying to fit both conventions in this package is too hard. Better to add a separate package (perhaps in the same git repo?) doing it the other way.

andrewthad:
Good to hear two people  say that. I have been thinking about it more this evening, and I  also think it’s  Best to stick with the mangled names.

andrewthad:
I'm about to push some more changes. I've tried to extend the naming convention that already in place, but there are a few ways in which I had to extend it. Here's the gist of it:

- Function names are suffixed with the type (`PrimArray`, `SmallArray`, etc.). For example: `foldlPrimArray`.
- Single-character suffixes (like `'` or `M`) are placed *after* the name of the type. For example: `foldlPrimArray'`, `imapPrimArrayM`.
- The suffix `P` (for `PrimBase`) is used as a more restrictive variant of `M` (for `Monad`). If you're working in `IO` or `ST`, there exists a faster implementation for traversals. Technically, in any sufficiently affine monad built on top of `IO` or `ST`, these should be able to work, but we do not have a typeclass (which would sit somewhere between `PrimMonad` and `PrimBase`) to discuss such types. Such work is left for a future discussions and a future PR. Example: `mapPrimArrayM` vs `mapPrimArrayP`.

Data/Primitive/PrimArray.hs
```diff
@@ -0,0 +1,488 @@
...
+    | otherwise = return (primArrayFromBackwardsListN sz xs)
+
+-- | Monadic map over a 'PrimArray' where the monad allows
+--   primitive actions.
+{-# INLINE mapPrimArrayP #-}
+mapPrimArrayP :: (PrimBase m, Prim a, Prim b)
+  => (a -> m b)
+  -> PrimArray a
+  -> m (PrimArray b)
+mapPrimArrayP = unsafeMapPrimArrayP
```
treeowl: I thought about trying for an appropriate constraint for the polymorphic arrays, but decided to just export the unsafe one. Whatever we do, we should do consistently. I believe the correct constraint is probably a type family-based one rather than a class-based one. We don't actually need to build a dictionary to represent the validity evidence; we just need to know that it's valid.

```haskell
class AffineEnoughF m => AffineEnough m
instance AffineEnoughF m => AffineEnough m

type family AffineEnoughF m :: ()
type instance AffineEnoughF (ST s) = '()
type instance AffineEnoughF IO = '()
type instance AffineEnoughF (ExceptT e m) = AffineEnoughF m
type instance AffineEnoughF (StateT m) = AffineEnoughF m
type instance AffineEnoughF (ReaderT e m) = AffineEnoughF m
type instance AffineEnoughF (ListT m) = TypeError ...
type instance AffineEnoughF (ContT r m) = TypeError ... -- I believe... need to check
```

Data/Primitive/PrimArray.hs
```diff
@@ -0,0 +1,488 @@
...
+  let !sz = sizeofPrimArray arr
+  marr <- newPrimArray sz
+  let go !ix = if ix < sz
+        then do
+          b <- f (indexPrimArray arr ix)
+          writePrimArray marr ix b
+          go (ix + 1)
+        else return ()
+  go 0
+  unsafeFreezePrimArray marr
```
treeowl: I think `traverse` rather than `map` explains more clearly what this does.

Data/Primitive/PrimArray.hs
```diff
@@ -0,0 +1,488 @@
...
+  let !sz = sizeofPrimArray arr
+  marr <- newPrimArray sz
+  let go !ix = if ix < sz
+        then do
+          let b = f (indexPrimArray arr ix)
+          writePrimArray marr ix b
+          go (ix + 1)
+        else return ()
+  go 0
+  unsafeFreezePrimArray marr
```
treeowl: I urge you to look at what I did with `runArray` in the polymorphic versions to elide the array boxes. It's simpler than it was before, and I think it should be done in this module too, and used in functions like this one.

Data/Primitive/PrimArray.hs
```diff
@@ -0,0 +1,488 @@
...
+  -> PrimArray a
+generatePrimArray len f = runST $ do
+  marr <- newPrimArray len
+  let go !ix = if ix < len
+        then do
+          writePrimArray marr ix (f ix)
+          go (ix + 1)
+        else return ()
+  go 0
+  unsafeFreezePrimArray marr
```
treeowl: You'll want applicative and prim-monadic versions of this too, won't you?

Data/Primitive/PrimArray.hs
```diff
@@ -0,0 +1,488 @@
...
+  -> PrimArray a
+  -> m (PrimArray b)
+mapPrimArrayM f arr = go 0 []
+  where
+  !sz = sizeofPrimArray arr
+  go !ix !xs
+    | ix < sz = do
+        !x <- f (indexPrimArray arr ix)
+        go (ix + 1) (x : xs)
+    | otherwise = return (primArrayFromBackwardsListN sz xs)
```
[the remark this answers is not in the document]
andrewthad: I like that approach (especially since it works for `Applicative` and not just `Monad`), but I was wondering about the performance difference between the two. I suspect that it builds up a closure on the heap, but I need to benchmark to see if I actually gain anything by building a list instead.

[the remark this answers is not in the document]
treeowl: Your way should work for `Applicative` too, I believe. I believe mine probably does *not* build a closure on the heap in common cases, and my bunchmarks came out well, but yes, you should benchmark for this rather different array type as well.

Data/Primitive/PrimArray.hs
andrewthad: I don't think we need a type family to do this. Could we not instead just have:

```
class PrimMonad m => PrimAffineMonad m where -- no methods
instance PrimAffineMonad m => PrimAffineMonad (MaybeT m)
instance PrimAffineMonad m => PrimAffineMonad (StateT s m)
... etc.
```

I don't see what advantage the type-family-based approach has over this.

Data/Primitive/PrimArray.hs
andrewthad: I guess `traverse` should really be preferred to `mapM` when naming things. I'll change this.

Data/Primitive/PrimArray.hs
treeowl: I just don't want to build any unnecessary evidence at runtime, but you may be right; your way may just make a pile of erased coercions. If your way makes good Core, I agree it's more pleasant.

Data/Primitive/PrimArray.hs
andrewthad: If we keep `generatePrimArray`, then yes. I wrote this, and then I started thinking about whether or not the array-constructing functions form `Data.Vector` (`generate`,`enumFromTo`,etc.) actually make sense here. The `primitive` package doesn't do fusion, so any use of `generatePrimArray` means that you're eating up extra space to represent something that is simply a function. Although, maybe there's a good use case around memoization. I think I'll keep it.

Data/Primitive/PrimArray.hs
andrewthad: Ah, yes. That does seem better. I'll switch to doing that.

Data/Primitive/PrimArray.hs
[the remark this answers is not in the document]
treeowl: Ah, you're talking about the delay in forcing till the end. A very valid concern, yes; that doesn't show up in the lazy traversals for polymorphic arrays. Maybe you *should* have both a `traverse` and a `mapM`. Worth benchmarking anyway.

Data/Primitive/PrimArray.hs
andrewthad: Actually, I'm conflicted about this. We potentially have four variants:

- mapping purely
- mapping under applicative context
- mapping under monadic context
- mapping under primitive monadic context

Names as it is now: `map`, `traverse`, `mapM`, `mapP`. I could change the latter two to `traverseM` and `traverseP`, but `traverseM` has no precedent.

andrewthad: I definitely agree with you on `map` and `traverse`. I'm afraid I don't understand the strictness-based distinction that you draw between the other two. For `PrimArray`, all four of these seem equally strict to me. If a single iteration produces a bottom, the whole thing always bottoms out.

treeowl: You missed it because it's subtle.

```haskell
traverse (maybe (pure undefined) print) $ fromList [Just 1, Nothing, Just 2, Just 3]
```

will print 1, 2, and 3 and then return `undefined`. The monadic versions will print 1 and then throw an exception.

andrewthad: Ah, thanks. That makes sense. I had completely missed that. Names could be the following: `map`, `traverse`, `traverse'`, `traverseP'`. Or we could just get rid of the monadic one and call the `PrimBase` one `traverse'`. After all, I think we could expect that someone who needs the extra performance is probably willing to go all the way.

treeowl: We're going to want to make the naming consistent with `Array` and `SmallArray`. Those have *two* notions of strict traversal:

1. Make sure all elements are forced when the resulting array is forced (this takes `Applicative`)
2. Force each element as it is produced (this takes `Monad`).

I'm not entirely sure that no one will want (1). In a sense, (2) seems to be redundant, since you can write

```haskell
foo :: (Traversable t, Monad m) => (a -> m b) -> t a -> m (t b)
foo f = traverse (\a -> f a >>= (pure $!))
```

but you may be able to be more efficient directly.

andrewthad: Possibly `unsafeTraversePrimArray'`. But my preference would be to just leave it unexported for the moment and then in a later PR, solve the affine prim monad problem with either a type class or a type family and then weaken the constraint on the `PrimBase` one. Then we wouldn't need the unsafe one.

treeowl: We'd still need the unsafe one. There are contexts where it will be perfectly fine. For example, as long as you're treating `Codensity m` sufficiently abstractly (with `liftCodensity` and `lowerCodensity`), it'll be just fine. I experimented with your way of doing the class, and to my surprise I wasn't able to make the constraints blow up. So we can go with that unless and until someone runs into a performance problem.

andrewthad: I've not ever actually used `Codensity`, but I'll take your word on that. For now, I'll go with:

- `mapPrimArray`
- `traversePrimArray` (Applicative)
- `traversePrimArray'` (PrimAffineMonad)
- `unsafeTraversePrimArray'` (PrimMonad)

The latter two functions are actually the same thing. I'll go ahead and add the methodless `PrimAffineMonad` class to make this work.

simonpj:
This ticket has a long thread that I have not been following carefully.   When you are done with it, would you like to broadcast a summary of the conclusion, and resulting changes, to a broader audience?  Thanks!

andrewthad:
@simonpj Once I implement the changes discussed above, I'll write up a summary and post it in a comment.

andrewthad:
@treeowl I was copying the approach used in `getSizeofMutableByteArray`. Whatever we do in one place should be done in the other. I think that the shimming that @bgamari used in https://github.com/haskell/primitive/pull/63 is pretty good. It results in no semantic difference between GHCs that can shrink byte arrays and ones that cannot.

treeowl:
Oh, right! I forgot it didn't get shrunk! You're absolutely right.

treeowl:
Could you add a comment in the source noting that in that version there really isn't any shrinking, so the test is safe? That'll keep anyone else from getting confused.

Data/Primitive/PrimArray.hs
treeowl: I'm not actually sure that `Codensity` is generally *unsafe* for this (it's a tad confusing), but I know a context where it definitely is safe. Anyway, exposing the unsafe version means people won't need orphan `PrimAffineMonad` instances.

andrewthad:
Alright, I think I've finally got a decent solution to the naming issue. For some background, in the inline comments above, @treeowl notes:

> mapPrimArray is a no-brainer. I think traversePrimArray makes sense for the Applicative version. The others (mapPrimArrayP and mapPrimArrayUnsafeP) are distinguished by their strictness: if an action successfully produces a bottom result, the operation is aborted. Perhaps the names should reflect strictness?

> We're going to want to make the naming consistent with Array and SmallArray. Those have two notions of strict traversal:
> 1. Make sure all elements are forced when the resulting array is forced (this takes Applicative)
> 2. Force each element as it is produced (this takes Monad).

There's more context in the actual comments, which I would encourage everyone to read. Here is the naming scheme for traversals and generators that I've tried to adhere to in the most recent commit:

- Strictness in elements: For `Array` and `SmallArray`, the suffix `'` (apostrophe) is used to indicate element-wise strictness. That is, all elements are forced once the array is pattern matched on. For `PrimArray` and `UnliftedArray`, this modifier is omitted since you cannot put a thunk either of these. Implicitly, every operator that produces a `PrimArray` or an `UnliftedArray` is understood to produce an array where everything has already been forced.
- Array-building strategy: For all array types, the suffix `P` is used to indicate that a more efficient strategy for building the array is being used. This requires `PrimAffineMonad` instead of `Applicative`, but it is always more efficient to use this when possible.

So, for example, the hypothetical `generate` functions for `Array` (not implemented in this PR) would look like this:

- `generateArray`: lazy in elements, non-effectful
- `generateArray'`: strict in elements, non-effectful
- `generateArrayA`: lazy in elements, requires `Applicative`
- `generateArrayP`: lazy in elements, requires `PrimAffineMonad`
- `generateArrayP'`: strict in elements, requires `PrimAffineMonad`
- `generateArrayUnsafe`: lazy in elements, requires `PrimMonad`, affinity unchecked
- `generateArrayUnsafe'`: strict in elements, requires `PrimMonad`, affinity unchecked

That's a lot of variants, but it does cover everything that people may need. Notably absent is `generateArrayA'`, which is not possible to write since we need `Monad` for this. Also, note that `generateArrayA` and `generateArrayP` are semantically identical. They should only differ in performance. For `PrimArray`, we would have:

- `generateArray`: strict in elements, non-effectful
- `generateArrayA`: strict in elements, but evaluates them to WHNF *after* executing all effects, requires `Applicative`
- `generateArrayP`: strict in elements, requires `PrimAffineMonad`
- `generateArrayUnsafe`: strict in elements, requires `PrimMonad`, affinity unchecked

I picked `generate` for this example because it makes the names easier. For `map`, we have `traverse` and `traverseP`. I would appreciate any feedback on this naming scheme.

treeowl:
Do we ever win anything from the `P'` variants? If the user has a `PrimMonad` on hand, can't they just do the forcing themselves? Is this just a small convenience? These other things are valuable because they enrich the safe interface. Does `P'` pull its weight? Caveat: my intuition around strictness is probably better than average, but definitely not perfect; I might have missed something significant.

andrewthad:
You're absolutely right. The `P'` and `Unsafe'` variants are unneeded. Removing them would shave two variants off of each of the `Array`-based traversals/generators.

andrewthad:
Just added `filterPrimArray`, which make pretty effective use of `resizeMutablePrimArray`.

andrewthad:
Also, here's an outline of what I plan on providing. I plan on implementing everything that be considered a map, a traversal, a fold, a way of generating an array. This includes effectful filtering and things like `mapMaybe` and its variants. I do not plan on implementing anything that deals with slicing. I also do not plan on implementing anything that requires the end user to handle an index into the array manually.

Again, I solicit any feedback. I'd like to make sure that I'm going down a path that others agree with.

andrewthad:
Just pushed a commit adding `filter` and `wither`. I borrowed the name `wither` from the `witherable` package. Do people like this name, or do people prefer `mapMaybe`?

RyanGlScott:
I prefer `mapMaybe`, since `mapMaybe` and `wither` have different types (`wither` is effectful, `mapMaybe` is not), and the type you've given corresponds to `mapMaybe`.

andrewthad:
Good catch. The function I wrote does not even correspond to `wither`.

Data/Primitive/PrimArray.hs
```diff
@@ -0,0 +1,817 @@
...
+foldlPrimArrayM' f z0 arr = go 0 z0
+  where
+    !sz = sizeofPrimArray arr
+    go !i !acc1
+      | i < sz = do
+          acc2 <- f acc1 (indexPrimArray arr i)
+          go (i + 1) acc2
+      | otherwise = return acc1
+
+-- | Traverse a primitive array. The traversal forces the resulting values and
```
cartazio: lets collapse traversePrimArrayP and unsafe... primitive doesn't provide any safety gurantees for back tracking land

cartazio:
Lets drop the AffinePrim array stuff, all of primimit is unsafe in the face of backtracking, we aren't going to add another class

treeowl:
Unsafe with respect to the property that no frozen array is mutated.

andrewthad:
I'm alright with dropping `PrimAffineMonad` for now. I still believe it would be useful, but it could always be discussed later and handled in a different PR if need be.

> When would anyone want to use PrimArray here rather than PrimVector / Unboxed vectors? Who is the customer/user that benefits?

Three different reasons stand out in my mind:

- `PrimArray` can be stored inside of `UnliftedArray`.
- `PrimArray`, when unpacked into a data constructor only uses a single machine word instead of three.
- With the NCG backend, `PrimArray` is faster than `Data.Vector.Primitive` . (With the LLVM backend, I am unable to observe a situation where `PrimArray` is faster). See the [benchmark that demonstrates this](https://gist.github.com/andrewthad/f7be4b3ac18c2c52be16d2e3c2bd6a97).

If you're building a patricia trie, `PrimArray` is going to be a better choice. You'll be able to use `UnliftedArray` to store prefixes, and you'll get fewer indirections. If you have a data type with several primitive vectors in it and if you don't need slicing, `PrimArray` will be a better choice. You can unpack more of these into a data constructor without causing the payload to exceed the size of a cache line.

Data/Primitive/PrimArray.hs
```diff
@@ -0,0 +1,796 @@
...
+      r -> isTrue# r
+#else
+      1# -> True
+      _ -> False
+#endif
+
+instance (Eq a, Prim a) => Eq (PrimArray a) where
+  a1@(PrimArray ba1#) == a2@(PrimArray ba2#)
+    | sameByteArray ba1# ba2# = True
+    | I# (sizeofByteArray# ba1#) /= I# (sizeofByteArray# ba2#) = False
```
treeowl: Don't we have a `sizeofByteArray`? If not, why not?

Data/Primitive/PrimArray.hs
```diff
@@ -0,0 +1,796 @@
...
+        go (a : as) !ix = if ix < len
+          then do
+            writePrimArray arr ix a
+            go as (ix + 1)
+          else die "fromListN" "list length greater than specified size"
+    go vs 0
+    unsafeFreezePrimArray arr
+
+primArrayToList :: forall a. Prim a => PrimArray a -> [a]
+primArrayToList = foldrPrimArray (:) []
```
treeowl: `primArrayToList xs = build (\c n -> foldrPrimArray c n xs)`, and probably mark it `INLINE`.

Data/Primitive/PrimArray.hs
andrewthad: We do. I've changed to use it insted.

Data/Primitive/PrimArray.hs
andrewthad: Done.

andrewthad:
Just fixed a cyclic import that I accidentally introduced in the previous commit.

andrewthad:
Ah,  the most recent push broke quickcheck-classes,  which includes a copy of prim array  and all of its associated functions.  I’ll need to push a new release of quickcheck-classes  to  make the test suite pass.  I’ll do that I am at a computer.

 That aside, is there any additional feedback?  If it would help make the pull request more palatable,  I could remove the  functions for traversing or constructing prim arrays.  They could always be discussed in a future PR instead.

RyanGlScott:
I'm afraid I feel woefully underqualified to review this, so I'm going to defer this to others.

@cartazio, what are your thoughts?

cartazio:
It’s on top of my oss review queue.  I’ve a few urgent work triage things I
need to progress this week before I do a close read through.

Please Hassle me over irc if I don’t get to it by the middle weekend :)

andrewthad:
poke

andrewthad:
[bother bother bother](https://www.youtube.com/watch?v=RomtlKowQs0)

andrewthad:
Motivational horse says "Review the PR."

```
                             ,(\_/)
                            ((((^`\
                          ((((  (6 \
                         ((((( ,    \
     ,,,_              (((((  /"._  ,`,
    ((((\\ ,...       ((((   /    `-.-'
    )))  ;'    `"'"'"((((    (
   (((  /           (((       \
    )) |                      |
   ((  |        .       '     |
   ))  \     _ '      `\   ,.'Y
   (   |   y;---,-""'"-.\   \/
   )   / ./  ) /         `\  \
      |./   ( (           / /'
      ||     \\          //'|
  jgs ||      \\       _// ||
      ||       ))     |_/  ||
      \_\     |_/          ||
      `'"                  \_\
                           `'"
```

treeowl:
Andrew, I suggest you try to convince Ryan to take a closer look. I know he
doesn't think he's qualified, but he is. Once this is merged, we need a
list of any changes that need to be made to the functions I added to Array
and SmallArray so those can be harmonized before release.

andrewthad:
@RyanGlScott I think you're qualified to review this. You reviewed my PR for `Data.Primitive.Ptr`, and this is basically the same thing but with more functions.

Data/Primitive/PrimArray.hs
```diff
@@ -0,0 +1,938 @@
...
+#if MIN_VERSION_base(4,7,0)
+import GHC.Exts (IsList(..))
+#endif
+
+#if MIN_VERSION_base(4,9,0)
+import Data.Semigroup (Semigroup)
+import qualified Data.Semigroup as SG
+#endif
+
+-- | Primitive arrays
```
RyanGlScott: The Haddocks here could stand to be spruced up some.

Data/Primitive/PrimArray.hs
```diff
@@ -0,0 +1,938 @@
...
+data PrimArray a = PrimArray ByteArray#
+
+-- | Mutable primitive arrays associated with a primitive state token
```
RyanGlScott: Similarly, this needs some sprucin'.

test/main.hs
```diff
@@ -84,10 +86,46 @@ main = do
       , lawsToTest (QCC.showReadLaws (Proxy :: Proxy (Array Int)))
 #if MIN_VERSION_base(4,7,0)
       , lawsToTest (QCC.isListLaws (Proxy :: Proxy ByteArray))
+#endif
+      ]
+    , testGroup "PrimArray"
+      [ lawsToTest (QCC.eqLaws (Proxy :: Proxy (PrimArray Word16)))
+      , lawsToTest (QCC.ordLaws (Proxy :: Proxy (PrimArray Word16)))
```
RyanGlScott: `PrimArray` has `Semigroup` and `Monoid` instances—should we test their properties here as well?

Data/Primitive/PrimArray.hs
```diff
@@ -0,0 +1,938 @@
...
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE MagicHash #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE UnboxedTuples #-}
+
+{-# OPTIONS_GHC -Wall #-}
+
+module Data.Primitive.PrimArray
```
RyanGlScott: There seems to be at least one set of non-standard naming conventions used in this module (e.g., the `-A` and `-P` suffix distinction), so perhaps it would be a good idea to give an overview of them in the module Haddocks.

Data/Primitive/PrimArray.hs
```diff
@@ -0,0 +1,938 @@
...
+    -- Here, we take the size in bytes, not in elements. We do this
+    -- since it allows us to defer performing the division to
+    -- calculate the size in elements.
+    sz1 = PB.sizeofByteArray (ByteArray ba1#)
+    sz2 = PB.sizeofByteArray (ByteArray ba2#)
+    loop !i
+      | i < 0 = True
+      | otherwise = indexPrimArray a1 i == indexPrimArray a2 i && loop (i-1)
+
+-- | __Note__: For the sake of efficiency, this is not a lexographic
```
RyanGlScott: s/lexographic/lexicographic

andrewthad: I cannot believe I've been misspelling this word my whole life.

andrewthad:
@RyanGlScott I've addressed all of your inline comments.

RyanGlScott:
Thanks, @andrewthad.

I'm going to be bold and land this. If there's other things we notice that need to be changed, we can do so in `master`.

PR #108, opened 2018-03-23
Data.Primitive.Ptr

andrewthad:
This PR provides functions for working with `Ptr` that take a `Prim` constraint instead of a `Storable` constraint. My motivation for this is that at work, I've ended up in situations where I've got a `Ptr` and a `MutableByteArray` referring to elements of the same type. It's weird to carry around both the `Storable` constraint and the `Prim` constraint in these situations especially because the two instances always agree with one another. To avoid the `Storable` constraint, I usually end up casting the `Ptr` to an `Addr`. This is boilerplate I would like to avoid, and it makes the whole process a little more error-prone.

Questions:

- Do other people think this is a good idea?
- Should these functions operate in `PrimMonad`, or should these be constrained to `IO`?

### DIFF
--- a/Data/Primitive/Ptr.hs
+++ b/Data/Primitive/Ptr.hs
@@ -1,0 +1,91 @@
+{-# LANGUAGE MagicHash #-}
+{-# LANGUAGE UnboxedTuples #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+-- |
+-- Module      : Data.Primitive.Ptr
+-- Copyright   : (c) Roman Leshchinskiy 2009-2012
+-- License     : BSD-style
+--
+-- Maintainer  : Roman Leshchinskiy <rl@cse.unsw.edu.au>
+-- Portability : non-portable
+--
+-- Primitive operations on machine addresses
+--
+
+module Data.Primitive.Ptr (
+  -- * Types
+  Ptr(..),
+
+  -- * Address arithmetic
+  nullPtr, advancePtr,
+
+  -- * Element access
+  indexOffPtr, readOffPtr, writeOffPtr,
+
+  -- * Block operations
+  copyPtr, movePtr, setPtr
+) where
+
+import Control.Monad.Primitive
+import Data.Primitive.Types
+
+import GHC.Base ( Int(..) )
+import GHC.Prim
+
+import GHC.Ptr
+import Foreign.Marshal.Utils
+
+
+-- | Offset a pointer by the given number of elements.
+advancePtr :: forall a. Prim a => Ptr a -> Int -> Ptr a
+advancePtr (Ptr a#) (I# i#) = Ptr (plusAddr# a# (i# *# sizeOf# (undefined :: a)))
+
+-- | Read a value from a memory position given by a pointer and an offset.
+-- The memory block the address refers to must be immutable. The offset is in
+-- elements of type @a@ rather than in bytes.
+indexOffPtr :: Prim a => Ptr a -> Int -> a
+{-# INLINE indexOffPtr #-}
+indexOffPtr (Ptr addr#) (I# i#) = indexOffAddr# addr# i#
+
+-- | Read a value from a memory position given by an address and an offset.
+-- The offset is in elements of type @a@ rather than in bytes.
+readOffPtr :: (Prim a, PrimMonad m) => Ptr a -> Int -> m a
+{-# INLINE readOffPtr #-}
+readOffPtr (Ptr addr#) (I# i#) = primitive (readOffAddr# addr# i#)
+
+-- | Write a value to a memory position given by an address and an offset.
+-- The offset is in elements of type @a@ rather than in bytes.
+writeOffPtr :: (Prim a, PrimMonad m) => Ptr a -> Int -> a -> m ()
+{-# INLINE writeOffPtr #-}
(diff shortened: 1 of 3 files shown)


Data/Primitive/Ptr.hs
```diff
@@ -0,0 +1,91 @@
...
+
+import GHC.Base ( Int(..) )
+import GHC.Prim
+
+import GHC.Ptr
+import Foreign.Marshal.Utils
+
+
+-- | Offset a pointer by the given number of elements.
+advancePtr :: forall a. Prim a => Ptr a -> Int -> Ptr a
```
RyanGlScott: This should probably be called `plusPtr` to match `plusAddr` in `Data.Primitive.Array`, but alas, `plusPtr`'s name was already taken by `base`. Oh well.

andrewthad:
The api is intended to mirror `Data.Primitive.Addr`. I didn't add `copyPtrToByteArray` because what I actually want is `copyPtrToPrimArray`, but I need to finish getting the `PrimArray` branch sorted out. I do need to add minus and rem. I'm trying to make sure I don't conflict with `base` on functions that do semantically different things. Calling the wrong `plusPtr` would be undetectable at compile time and would have devastating effects at runtime, so I instead chose to use the name `advancePtr`, whose doppelganger in `Foreign.Marshal.Alloc` has identical semantics. And `minusPtr` has a similar issue, but there's nothing in base that does an element-wise pointer subtraction. Maybe `subtractPtr` would be alright.

RyanGlScott:
To be clear: I agree with your naming choices. My objections were more rhetorical than anything :)

In that case, if you add `subtractPtr` and `remPtr`, then I think we can declare victory here.

andrewthad:
I've added `subtractPtr` and made some other corrections. After thinking about it more, I'm not sure that `remPtr` can be given a good meaningful definition. If we have `myPtr :: Ptr Int64`, then the expression `remPtr myPtr 8` should always be zero. If we do `remPtr myPtr 32`, then we'll always get something between 0 and 3. Normally, one would expect that `customRemainderFunction x n` would return something between in the range `[0,n - 1]`, so may be an unexpected behavior.

There's an alternative definition that makes more sense. We could first divide the pointer by the element size (which should always divide evenly into it) and then call `rem`. The results of such a function would match most users' intuitions concerning what a `rem` function should do.

Regardless, the question arises: why would anyone use this? People use `remAddr` to round to the nearest aligned `Ptr`. But, as the argument to `remPtr` would already be an aligned `Ptr`, what would we be doing with this function? Are we trying to get additional alignment? (maybe for AVX register loading)

Anyway, let me know which implementation, if any, you think is more appropriate, and I'll add it.

Data/Primitive/Ptr.hs
```diff
@@ -0,0 +1,99 @@
...
+{-# INLINE readOffPtr #-}
+readOffPtr (Ptr addr#) (I# i#) = primitive (readOffAddr# addr# i#)
+
+-- | Write a value to a memory position given by an address and an offset.
+-- The offset is in elements of type @a@ rather than in bytes.
+writeOffPtr :: (Prim a, PrimMonad m) => Ptr a -> Int -> a -> m ()
+{-# INLINE writeOffPtr #-}
+writeOffPtr (Ptr addr#) (I# i#) x = primitive_ (writeOffAddr# addr# i# x)
+
+-- | Copy the given number of bytes from the second 'Ptr to the first. The
```
RyanGlScott: You're missing an apostrophe at the end of 'Ptr.

Also, that should read "Copy the given number of _elements_", not "bytes", yes?

Data/Primitive/Ptr.hs
```diff
@@ -0,0 +1,99 @@
...
+copyPtr :: forall m a. (PrimMonad m, Prim a)
+  => Ptr a -- ^ destination pointer
+  -> Ptr a -- ^ source pointer
+  -> Int -- ^ number of elements
+  -> m ()
+{-# INLINE copyPtr #-}
+copyPtr (Ptr dst#) (Ptr src#) n
+  = unsafePrimToPrim $ copyBytes (Ptr dst#) (Ptr src#) (n * sizeOf (undefined :: a))
+
+-- | Copy the given number of bytes from the second 'Ptr to the first. The
```
RyanGlScott: Same comments as in `copyPtr`.

andrewthad:
Excellent. Let's leave out `remPtr`. I've made the corrections you provided.

Data/Primitive/Ptr.hs
```diff
@@ -0,0 +1,99 @@
...
+{-# INLINE readOffPtr #-}
+readOffPtr (Ptr addr#) (I# i#) = primitive (readOffAddr# addr# i#)
+
+-- | Write a value to a memory position given by an address and an offset.
+-- The offset is in elements of type @a@ rather than in bytes.
+writeOffPtr :: (Prim a, PrimMonad m) => Ptr a -> Int -> a -> m ()
+{-# INLINE writeOffPtr #-}
+writeOffPtr (Ptr addr#) (I# i#) x = primitive_ (writeOffAddr# addr# i# x)
+
+-- | Copy the given number of bytes from the second 'Ptr' to the first. The
```
RyanGlScott: Should this read "elements" instead of "bytes"?

PR #109, opened 2018-03-25
Heterogeneous array creation

### LINKED ISSUE #103: Offer an extremely general combinator for building arrays
This may look a bit strange....

```haskell
create :: (forall h f g.
            Applicative h => (forall x. f x -> h (g x)) -> t f -> h (t g))
       -> (forall s. ST s (t (MutableArray s)))
       -> t Array
create f m = runST $ m >>= f unsafeFreezeArray
```

Using something like `Vinyl`, users can use this function to use a single `ST` action to safely produce multiple arrays, of multiple types, along with arbitrary additional context. It's much easier to understand with the [Traversable](https://hackage.haskell.org/package/rank2classes-1.0.2/docs/Rank2.html#t:Traversable) class from `rank2classes`:

```haskell
runArraysHet :: Rank2.Traversable t
             => (forall s. ST s (t (MutableArray s)))
             -> t Array
runArraysHet m = runST $ m >>= Rank2.traverse unsafeFreezeArray
```

but `primitive` obviously has no business depending on `rank2classes`.

treeowl:
Create arbitrarily many arrays of arbitrarily many types from one
`ST` action.

Closes #103


treeowl:
Question: I made the heterogeneous creation functions take traversals that are polymorphic over their `Applicative`s. Is that necessary to ensure the safety of the `unsafeFreeze` calls, or would it be okay to specialize to `ST s`? The rest of the polymorphism seems to be necessary.

treeowl:
It might be worth adding a little more flexibility to the type of `runSmallArraysHetOf` to avoid extra traversals. This is an extreme version:

```haskell
runArraysHetOfFull
  :: Applicative q
  => (forall h f g.
       (Applicative h => (forall x. f x -> h (g x)) -> t f -> h (u g))) -- ^ A rank-2 traversal
  -> (forall x. Array x -> q (r x))  -- ^ What to traverse over the final container of arrays
  -> (forall s x. p s x -> MutableArray s x) -- ^ How to get mutable arrays from the ST result
  -> (forall s. ST s (t (p s))) -- ^ An 'ST' action producing a rank-2 container
  -> q (u r)
runArraysHetOfFull trav post pre m = runST $ m >>= getCompose . trav (Compose . fmap post . unsafeFreezeArray . pre)
```

treeowl:
FYI, the higher-rank traversals are basically the ones from `Rank2.Traversable` in `rank2classes` and from `Data.Comp.Multi.HTraversable` in `compdata`. They're type-flipped from `Data.Vinyl.Core.rtraverse` in `vinyl` (we definitely don't want it flipped like that). The only difference is that the traversals can be *type-changing*. I don't think that *likely* lets anyone break things (though I haven't proven that), and it seems to be somewhat useful.

Data/Primitive/SmallArray.hs
```diff
@@ -940,3 +947,74 @@ smallArrayFromListN n l = SmallArray (Array.fromListN n l)
...
+-- @'Compose' ('Either' e) (c,)@. For a more general version, see
+-- 'runArraysHetOf'.
+runSmallArrays
+  :: Traversable t
+  => (forall s. ST s (t (SmallMutableArray s a)))
+  -> t (SmallArray a)
+runSmallArrays m = runST $ m >>= traverse unsafeFreezeSmallArray
+
+-- | Create arbitrarily many arrays that may have different types. For
+-- a simpler but less general version, see 'runArrays'.
```
andrewthad: Should say "see `runSmallArrays`".

andrewthad:
I cannot verify the soundness of this, but I like it. Also, it would be good to know the answer of your question about `Applicative` possibly being specialized to `ST` instead.

andrewthad:
Also, I dislike the names with Het in them,  but at the moment, I don’t have any better suggestions.

treeowl:
The names are awful.

Data/Primitive/Array.hs
```diff
@@ -1,6 +1,9 @@
 {-# LANGUAGE CPP, MagicHash, UnboxedTuples, DeriveDataTypeable, BangPatterns #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE TypeFamilies #-}
+#if __GLASGOW_HASKELL__ >= 708
```
RyanGlScott: Shouldn't this be 706? (That's the first GHC version that reliably supported `PolyKinds`.)

Data/Primitive/SmallArray.hs
```diff
@@ -7,6 +7,9 @@
 {-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE BangPatterns #-}
+#if __GLASGOW_HASKELL__ >= 708
```
RyanGlScott: Similarly, 706 here.

treeowl:
Using `Data.Functor.Sum`, I've convinced myself that the type change from `t` to `u` shouldn't be a problem. I'm still not so sure about whether we can have trouble with unlawful traversals in general. I think the very *most* we might be able to get away with is this:

```haskell
runArraysHetOfThen
  :: Applicative q
  => (forall s1 s2.
       ((forall x. MutableArray s2 x -> Compose (ST s1) q (r x)) -> t (MutableArray s2) -> Compose (ST s1) q (u r)))
  -> (forall x. Array x -> q (r x))
  -> (forall s. ST s (t (MutableArray s)))
  -> q (u r)
```

This reveals *almost* everything to the traversal function, but doesn't give it quite enough to actually monkey with the `MutableArray`s itself (since it doesn't know it's working in the same state thread they live in).

treeowl:
I've pretty much convinced myself that these are safe, even generalized considerably. See [my answer to my question on SO](https://stackoverflow.com/a/49836335/1477667).

Data/Primitive/Array.hs
treeowl: I mistakenly though it didn't stabilize till later. Fixed!

RyanGlScott:
Do you mind including a summary of the rationale in that SO post somewhere in the comments? Once that's included, I'd say this is ready to be merged.

treeowl:
@RyanGlScott, I added some notes, changed some types, and removed `PolyKinds` (I don't think it actually does anything useful here). Could you take another glance?

treeowl:
Oh, and a note I meant to leave earlier:

I've generalized the types of `runArraysHetOf` and `runSmallArraysHetOf`. Thanks to that generalization, `runArraysHetOfThen` and `runSmallArraysHetOfThen` can be implemented (efficiently) using them, so I've removed the latter two functions to reduce clutter. I've *added* `runArraysOf` and `runSmallArraysOf` for consistency and because not everyone who might want to supply a custom traversal will also want to figure out rank-2 traversals.

treeowl:
One more question: would `runHetArraysOf` roll more easily off the tongue?

treeowl:
I think it would; if other's don't disagree, I'll make that change.

cartazio:
I’ll have a read through after I finish my other bits of review work

Data/Primitive/Array.hs
```diff
@@ -798,3 +799,180 @@ instance (Typeable s, Typeable a) => Data (MutableArray s a) where
...
+--      -- ^ An 'ST' action producing a rank-2 container of 'MutableArray's.
+--   -> q u
+-- runHetArraysOfThen trav post m = getConst $
+--   runHetArraysOf (\f -> coerce . getCompose . trav (Compose . fmap post . f)) m
+-- @
+runHetArraysOf
+  :: (forall s1 s2.
+       ((forall x. MutableArray s1 x -> ST s2 (Array x)) -> t (mut s1) -> ST s2 u))
+     -- ^ A rank-2 traversal
+  -> (forall s. ST s ((t :: (* -> *) -> *) (mut s)))
```
treeowl: With `PolyKinds`, this can be `t :: (j -> k) -> *`. Maybe we can do that once we drop support for 7.4; until then I think it's probably a bit too much CPP for what it offers.

treeowl: (Remember that `mut` doesn't necessarily have to be used at all; it's just there to make the type match that of `htraverse`. So if someone wants to use a container of `MutableArray#`, say, they can still do that with a bit of type trickery.)

treeowl:
I deleted my last comment because it was a bit wrong, thankfully.

I've been thinking some more, and there is another heterogeneous option:

```haskell
class Freezable m i | m -> i where
  type GetS m :: *
  uf :: m -> ST (GetS m) i

instance Freezable (MutableByteArray s) ByteArray where
  type GetS (MutableByteArray s) = s
  uf = unsafeFreezeByteArray

instance a ~ b => Freezable (MutableArray s a) (Array b) where
  type GetS (MutableArray s a) = s
  uf = unsafeFreezeArray

instance a ~ b => Freezable (SmallMutableArray s a) (SmallArray b) where
  type GetS (SmallMutableArray s a) = s
  uf = unsafeFreezeSmallArray

runVariousHetArraysOf
  :: (forall s1 s2.
       ((forall m i. (Freezable m i, GetS m ~ s1) => m -> ST s2 i) -> t (mut s1) -> ST s2 u))
  -> (forall s. ST s ((t :: k -> *) (mut s)))
  -> u
runVariousHetArraysOf trav m = runST $ m >>= trav uf
```

The nice thing is that you can run one `ST` action and get all *sorts* of different results all at once: `Array`s, `SmallArray`s, `ByteArray`s, etc. Since the `Freezable` class is open, it could easily be instantiated for types living in other packages.

treeowl:
Aha! I just realized something. If I modify `runArraysOf` very slightly, I can actually implement `runHetArraysOf` on top of it! Specifically, if I change the type signature of `runArraysOf` to

```haskell
runArraysOf
  :: (forall s1 s2.
       (forall x. MutableArray s1 x -> ST s2 (Array x)) -> t (mut s1 a) -> ST s2 u)
  -> (forall s. ST s (t (mut s a)))
  -> u
```

then I can write

```haskell
runHetArraysOf trav m = runArraysOf (\f -> trav f . unT) $ T <$> m

newtype T (t :: (* -> *) -> *) (msa :: *) = T {unT :: t (Fun msa)}
type family Fun x where
  Fun (f x) = f
```

cartazio:
One thing I’ve been thinking about is maybe this set of combinators would
make more sense to live in lens.  A lot of the utility seems like it’d be
highest for folks who use lense a lot.  Especially the whole higher rank
traversal bit.

I guess what I mean is that comfort using these combinators seems
predicated on using lens, which is down stream of primitive.

At a philosophical level, my perspective on the primitive package is that
it’s about making the various magic hash  mutable data structures ghc
provides usable from types in kind Star. Vector and similar higher level
libraries  are the clients / users of primitive for their data structure
needs


The only exception is  We can’t do that safely for STM Data structures in
primitive because of the no nesting transactions requirement of stm
precluding execution in any setting / prim monad aside from those with io
as the base monad.

So I guess what I’m trying to say is that I’m currently inclined to say no
to adding these traversal combinators to primitive.  Though I’ll ponder
this a bit more before I take a strong stance

treeowl:
That might be reasonable. On the other hand, one of the versions I
described uses a Freezable class to be able to freeze anything with a
notion of freezing. That Freezable class, to be maximally useful, should
live at the lowest level of the module hierarchy that it can. Optimally,
that would probably mean GHC.Arr, but starting it somewhere in primitive
might be a good compromise. With that class available, anyone can write a
Trustworthy version of runArraysOf with their desired type details.

treeowl:
Let met take that back a bit... The ergonomics of the extreme class-based
version don't seem that great. Maybe I can fix that, but maybe not. The
array-type-specific versions, on the other hand, have great ergonomics, and
there doesn't seem to be anywhere more sensible to put them than here.

andrewthad:
In my mind, `runArrays` is clearly useful and should be included. I doubt that I personally would ever remember to use anything beyond `runArrays`. I would probably forget and just do it without the combinators. Here are my general documentation-related concerns:

- What is a rank-2 traversal? Are there laws that such a function must satisfy? A google search for "haskell rank 2 traversal" turns up nothing relevant. The docs should probably link to resources or other haskell libraries.
- An example of how to use `runHetArraysOf` is provided. This is good. But, is this function intended to be used with functions from some other library? Are the arguments taken in a particular order for a particular reason? Is this intended to be used with `lens` or something like `lens`? If so, it would be nice to include the lens-type-synonym translation of the type signature in the docs.

My other concern is that we have the following implementations:

```
runArraysOf trav m = runST $ m >>= trav unsafeFreezeArray
runHetArraysOf trav m = runST $ m >>= trav unsafeFreezeArray
```

These two functions have identical implementations and slightly different type signatures. It bothers me that the type signatures of these two cannot somehow be unified, but there seems to be no way around this.

Also, there's a annoying asymmetry going on. We don't have a `runHetArrays` function (with a `TraversableRank2` typeclass constraint or whatever the appropriate typeclass is named). This makes me more inclined to say that `runHetArraysOf` belongs in some other package, but it does seem to be pretty useful in the example given, so I guess I like it here.

cartazio:
These are valid points, but let’s make sure we don’t scope creep the role
of the primitive package :)

PR #114, opened 2018-03-28
Add benchmark suite

andrewthad:
Benchmark performance difference between closure-building array traversal and the unsafe PrimMonad variant. Currently, `cabal new-build` fails to build this if you are using GHC 8.0 or newer because of something going on in `basement` (which `gauge` depends on). I'm sure this will be resolved soon. For now, we can do:

```
$ cabal new-build bench --enable-benchmarks -w ghc-7.10.3
$ ./dist-newstyle/build/x86_64-linux/ghc-7.10.3/primitive-benchmarks-0.1/c/bench/build/bench/bench
benchmarked Array/traverse/closure
time                 929.5 μs   (913.6 μs .. 948.6 μs)
                     0.998 R²   (0.998 R² .. 0.999 R²)
mean                 910.4 μs   (906.5 μs .. 916.2 μs)
std dev              15.65 μs   (10.99 μs .. 20.36 μs)

benchmarked Array/traverse/unsafe
time                 360.6 μs   (357.1 μs .. 363.8 μs)
                     1.000 R²   (0.999 R² .. 1.000 R²)
mean                 354.3 μs   (353.1 μs .. 355.6 μs)
std dev              4.406 μs   (3.300 μs .. 6.144 μs)
```

This helps confirm the suspicion that the unsafe variant is actually faster. I've measured these two functions by copying their implementations into the benchmarks directory. For benchmarking multiple possible implementations of the same thing, I prefer copying code into the benchmark suite like this. In the event that we ever change the implementation of `traverse`, we will still be able to measure it against the other options.

### DIFF
--- a/bench/README.md
+++ b/bench/README.md
@@ -1,0 +1,7 @@
+Benchmark Suite
+=======================
+
+The benchmark suite for `primitive` cannot be included in the same package
+as `primitive` itself. The benchmark suite depends on `gauge`, which
+transitively depends on `primitive`. To break up this dependency cycle,
+the test suite lives here in its own unpublished package.
--- a/.travis.yml
+++ b/.travis.yml
@@ -93,7 +93,7 @@ install:
       cabal new-update head.hackage -v
     fi
   - grep -Ev -- '^\s*--' ${HOME}/.cabal/config | grep -Ev '^\s*$'
-  - "printf 'packages: \".\" \"./test\"\\n' > cabal.project"
+  - "printf 'packages: \".\" \"./test\" \"./bench\"\\n' > cabal.project"
   - echo 'package quickcheck-classes' >> cabal.project
   - "echo '  flags: -aeson -semigroupoids' >> cabal.project"
   - cat cabal.project
@@ -103,8 +103,11 @@ install:
   - if [ -f "./test/configure.ac" ]; then
       (cd "./test" && autoreconf -i);
     fi
+  - if [ -f "./bench/configure.ac" ]; then
+      (cd "./bench" && autoreconf -i);
+    fi
   - rm -f cabal.project.freeze
-  - rm -rf .ghc.environment.* "."/dist "./test"/dist
+  - rm -rf .ghc.environment.* "."/dist "./test"/dist "./bench"/dist
   - DISTDIR=$(mktemp -d /tmp/dist-test.XXXX)
 
 # Here starts the actual work to be performed for the package under test;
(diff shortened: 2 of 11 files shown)


bench/README.md
```diff
@@ -0,0 +1,7 @@
+Benchmark Suite
+=======================
+
+The benchmark suite for `primitive` cannot be included in the same package
+as `primitive` itself. The benchmark suite depends on `gauge`, which
+transitively depends on `primitive`. To break up this dependency cycle,
+the test suite lives here in its own unpublished package.
```
RyanGlScott: s/the test suite/the benchmark suite

andrewthad:
I had not considered that travis would build the benchmark suite, but that's probably a good idea. I wonder if it's important that the benchmark suite be able to build with old versions of GHC. What if we only built the benchmark suite with the latest two versions of GHC instead? Do we lose anything by doing that?

The difficulty with benchmark libraries is that there's basically just `criterion`, which has a ton of dependencies, and `gauge`, which is new and stripped down but depends on `basement` (which makes me a little wary, possibly without reason). The third option would be to roll our own, but I'm not keen on doing that.

That was all a bit of a ramble. I think what I'm trying to say is that I would rather have a benchmark suite that's occasionally gets broken or doesn't build with old GHCs than not having one at all. What do you think?

RyanGlScott:
I would also be fine with only supporting the benchmark suite on a limited number of GHC versions. That should work fine with Travis too—we should be able to engineer things such that `cabal new-build` doesn't attempt to build the benchmark suite if `base` isn't sufficiently new enough.

Obviously, the more GHCs we can benchmark, the better. But it needn't necessarily be a bottleneck.

andrewthad:
I realized that I forgot to answer your other question. Yes, the goal would be to also benchmark the actual implementation in use. I'll add a few benchmarks for some other functions that are actually exported by `primitive` itself and place them in their own section.

treeowl:
FWIW, my philosophy for `containers` has generally been that it's not worth the trouble and CPP to try to make things snappy on older GHC versions. That way lies lots of work and short-lived, bitter-sweet rewards.

RyanGlScott:
I've just pushed a commit which should permit building the benchmark suite on Travis when using GHC 7.10 or later (since that is, AFAICT, the oldest version of GHC that `gauge` currently supports).

RyanGlScott:
There appear to be various build failures for `quickcheck-classes` on versions of GHC older than 7.10. @andrewthad, do you know what is going on here?

andrewthad:
Yes, I just noticed this on the prim array branch. I'm working on it.

andrewthad:
@RyanGlScott If you bump to requiring `quickcheck-classes-0.4.8`, it should be working now. I had accidentally introduced a regression.

andrewthad:
Woah. I did not realize that github would let you push to branches in forks that had open PRs. That's so great.

treeowl:
@RyanGlScott, you should know that `containers` is dropping support for GHC 7.4. The next non-point release will only support 7.6 and above.

andrewthad:
@RyanGlScott I changed the `enable-benchmarks` flag to be manual (I don't really understand what that means though), but now it's making travis fail. I can change it back, but I cannot figure out how to bench the benchmark suite locally when the flag isn't manual. Could you show me what command you run locally to build it?

RyanGlScott:
I'm not sure how you're trying to build this, but on the previous commit, I could build `primitive` plus the benchmarks like so:

```
$ cabal new-build all --enable-benchmarks --dry
Resolving dependencies...
Build profile: -w ghc-8.4.1 -O1
In order, the following would be built (use -v for more details):
 - primitive-0.6.3.0 (lib) (first run)
 - vector-0.12.0.1 (lib) (requires build)
 - gauge-0.2.1 (lib) (requires build)
 - primitive-benchmarks-0.1 (bench:bench) (first run)
```

Having that flag not be manual is important because otherwise, the `cabal-install` solver won't clue in to the fact that it shouldn't attempt to build the benchmark suite on older versions of GHC (since it's guarded behind a `Cabal` flag).

andrewthad:
That's funny. I must have messed up my build environment earlier. The command you gave works. Thanks.

RyanGlScott:
Is this ready to go, or are you still planning further changes?

andrewthad:
There are some more benchmarks that I would like to add eventually, but  I would say that it is ready to go as is. It probably makes more sense to add the additional benchmarks in subsequent PRs.

RyanGlScott:
Alright, I'll opt to merge this, then. Thanks, @andrewthad!

Issue #115, opened 2018-03-30
shim for resizeMutableByteArray is wrong


andrewthad:
While testing the functions for filtering a `PrimArray`, I discovered that the implementation of `resizeMutablePrimArray` behaved incorrectly on old GHCs. I had copied this from `resizeMutableByteArray`. For old GHCs, the shim currently looks like this:

```
resizeMutableByteArray arr n
  = do arr' <- newByteArray n
       copyMutableByteArray arr 0 arr' 0 (min (sizeofMutableByteArray arr) n)
       return arr'
```

The call to `copyMutableByteArray` has the source and destination arrays flipped.

treeowl:
Reopening to add a test.

andrewthad:
I'm waiting on https://github.com/haskell/primitive/pull/118 before adding a test. That branch provides `byteArrayFromList` and `byteArrayToList`, which make it much easier to write tests for `ByteArray` that work with all GHC versions (since `IsList` doesn't exist on older GHCs).

andrewthad:
I have written a test: https://github.com/haskell/primitive/pull/128

PR #116, opened 2018-03-30
correct the shim for resizeMutableByteArray

andrewthad:
Resolves https://github.com/haskell/primitive/issues/115


treeowl:
Good catch. It would be good to add a test.

Issue #113, opened 2018-03-27
Wrappers for isMutableByteArrayPinned# and isByteArrayPinned#


andrewthad:
We should introduce wrapper functions for `isMutableByteArrayPinned#` and `isByteArrayPinned#`. They would return `Bool`. I don't think these can be shimmed for GHCs that don't have these primops.

PR #117, opened 2018-03-30
implement isByteArrayPinned and isMutableByteArrayPinned

andrewthad:
Resolves https://github.com/haskell/primitive/issues/113.

### DIFF
--- a/Data/Primitive/ByteArray.hs
+++ b/Data/Primitive/ByteArray.hs
@@ -41,7 +41,11 @@ module Data.Primitive.ByteArray (
   -- * Information
   sizeofByteArray,
   sizeofMutableByteArray, getSizeofMutableByteArray, sameMutableByteArray,
+#if __GLASGOW_HASKELL__ >= 802
+  isByteArrayPinned, isMutableByteArrayPinned,
+#endif
   byteArrayContents, mutableByteArrayContents
+
 ) where
 
 import Control.Monad.Primitive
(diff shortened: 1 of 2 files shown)


RyanGlScott:
Looks good to me! Can you also add an entry to the changelog?

andrewthad:
I've really got to get changelogs integrated into my workflow better. After a lifetime of not doing them, the habit is still weak. Thanks for the reminder.

Data/Primitive/ByteArray.hs
```diff
@@ -192,6 +200,26 @@ sizeofMutableByteArray :: MutableByteArray s -> Int
...
+#if __GLASGOW_HASKELL__ >= 802
+-- | Check whether or not the byte array is pinned. Pinned byte arrays cannot
+--   be moved by the garbage collector. It is safe to use 'byteArrayContents'
+--   on such byte arrays. This function is only available when compiling with
+--   GHC 8.2 or newer.
+isByteArrayPinned :: ByteArray -> Bool
+{-# INLINE isByteArrayPinned #-}
+isByteArrayPinned (ByteArray arr#) = case Exts.isByteArrayPinned# arr# of
+  1# -> True
+  _ -> False
```
treeowl: This should probably use `GHC.Exts.isTrue#`

andrewthad:
I doubt that the `INLINE` pragmas give us anything in this case. GHC will certainly decide to inline a function this small on its own. I only put them there for consistency. Every other function here has an `INLINE` pragma.

Issue #107, opened 2018-03-23
Use compareByteArrays# on GHC 8.4+


andrewthad:
On GHC 8.4 and newer, we have [compareByteArrays#](https://www.stackage.org/haddock/nightly-2018-03-23/ghc-prim-0.5.2.0/GHC-Prim.html#v:compareByteArrays-35-). It has a [pointer equality short cut built in](https://phabricator.haskell.org/rGHC31c260f3967d2c06063c962a98475058daa45c6d), so there's no need for `sameByteArray`. Plus, it doesn't use the FFI, so it should be a little bit faster.

andrewthad:
Whoops, forgot to mention where I think this should be used: the `Eq` and `Ord` instances for `ByteArray`.

treeowl:
Will the semantics remain the same?

andrewthad:
They should since `compareByteArrays#` is just `memcmp` under the hood. We should probably add a property test for that just to make sure though.

treeowl:
Doesn't `ByteArray` have a weird `Ord` instance?

andrewthad:
It does. But the behavior of the instance would be unchanged.

hvr:
Fwiw, here's the compat-shim I use in `text-short`: https://hackage.haskell.org/package/text-short-0.1.2/src/src-ghc708/PrimOps.hs

andrewthad:
@hvr Thanks for the shim. I ended up tweaking it a little for `ByteArray`'s more narrow needs, but it was a helpful starting point.

I'm putting up a PR for this.

PR #118, opened 2018-03-30
Use compareByteArrays#

andrewthad:
Resolves https://github.com/haskell/primitive/issues/107

Use compareByteArrays`  for the Eq and Ord instance of ByteArray when building with GHC 8.4 or newer. Add tests to ensure that the behavior of the Eq and Ord instances are consistent across different versions of GHC. Additionally, export byteArrayFromList and byteArrayFromListN, which are helpful for making the test suite backwards-compatible.

Hold off on merging this until I get the benchmark suite in better shape. I want to ensure that this doesn't somehow introduce a performance regression.

### DIFF
--- a/Data/Primitive/ByteArray.hs
+++ b/Data/Primitive/ByteArray.hs
@@ -24,6 +24,9 @@ module Data.Primitive.ByteArray (
   -- * Element access
   readByteArray, writeByteArray, indexByteArray,
 
+  -- * Constructing
+  byteArrayFromList, byteArrayFromListN,
+
   -- * Folding
   foldrByteArray,
 
@@ -63,7 +66,6 @@ import Data.Typeable ( Typeable )
 import Data.Data ( Data(..) )
 import Data.Primitive.Internal.Compat ( isTrue#, mkNoRepType )
 import Numeric
-import System.IO.Unsafe
 
 #if MIN_VERSION_base(4,9,0)
 import qualified Data.Semigroup as SG
@@ -72,6 +74,12 @@ import qualified Data.Foldable as F
 
 #if !(MIN_VERSION_base(4,8,0))
 import Data.Monoid (Monoid(..))
+#endif
+
+#if __GLASGOW_HASKELL__ >= 804
+import GHC.Exts (compareByteArrays#)
+#else
+import System.IO.Unsafe (unsafeDupablePerformIO)
 #endif
 
 -- | Byte arrays
@@ -223,17 +231,20 @@ foldrByteArray f z arr = go 0
       | otherwise                    = z
     sz = sizeOf (undefined :: a)
 
-fromListN :: Prim a => Int -> [a] -> ByteArray
-fromListN n ys = runST $ do
+byteArrayFromList :: Prim a => [a] -> ByteArray
+byteArrayFromList xs = byteArrayFromListN (length xs) xs
+
+byteArrayFromListN :: Prim a => Int -> [a] -> ByteArray
+byteArrayFromListN n ys = runST $ do
     marr <- newByteArray (n * sizeOf (head ys))
     let go !ix [] = if ix == n
           then return ()
-          else die "fromListN" "list length less than specified size"
+          else die "byteArrayFromListN" "list length less than specified size"
         go !ix (x : xs) = if ix < n
           then do
             writeByteArray marr ix x
             go (ix + 1) xs
-          else die "fromListN" "list length greater than specified size"
+          else die "byteArrayFromListN" "list length greater than specified size"
     go 0 ys
     unsafeFreezeByteArray marr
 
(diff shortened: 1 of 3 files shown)


treeowl:
Do you know how `compareByteArrays#` is implemented? If it uses the FFI, it could be slower for array pairs with sufficiently short common prefixes. But I think we should probably do it regardless.

Data/Primitive/ByteArray.hs
```diff
@@ -223,17 +231,20 @@ foldrByteArray f z arr = go 0
       | otherwise                    = z
     sz = sizeOf (undefined :: a)
 
-fromListN :: Prim a => Int -> [a] -> ByteArray
-fromListN n ys = runST $ do
+byteArrayFromList :: Prim a => [a] -> ByteArray
+byteArrayFromList xs = byteArrayFromListN (length xs) xs
```
treeowl: We have shrinking here, so we should probably benchmark an array doubling (or similar) option. Array doubling would let this be a "good consumer" for list fusion, as well as compacting the representation much more aggressively regardless.

Data/Primitive/ByteArray.hs
```diff
@@ -223,17 +231,20 @@ foldrByteArray f z arr = go 0
...
-          else die "fromListN" "list length less than specified size"
+          else die "byteArrayFromListN" "list length less than specified size"
         go !ix (x : xs) = if ix < n
           then do
             writeByteArray marr ix x
             go (ix + 1) xs
-          else die "fromListN" "list length greater than specified size"
+          else die "byteArrayFromListN" "list length greater than specified size"
     go 0 ys
     unsafeFreezeByteArray marr
```
treeowl: At some point, this should probably use a `runByteArray` or `createByteArray` here.

test/main.hs
```diff
@@ -78,6 +83,20 @@ main = do
...
+
+byteArrayEqProp :: QC.Property
+byteArrayEqProp = QC.property $ \(xs :: [Word8]) (ys :: [Word8]) ->
+  (compareLengthFirst xs ys == EQ) === (byteArrayFromList xs == byteArrayFromList ys)
+
+compareLengthFirst :: [Word8] -> [Word8] -> Ordering
+compareLengthFirst xs ys = case compare (length xs) (length ys) of
+  LT -> LT
+  GT -> GT
+  EQ -> compare xs ys
```
treeowl: It might be more idiomatic to write

```haskell
compareLengthFirst xs ys = (compare `on` length) xs ys <> compare xs ys
```

andrewthad: That's certainly more concise. I'll switch to that.

andrewthad:
I don't know how `compareByteArrays#` is implemented. If it uses the FFI, then it's certainly no worse than what we are doing now. I think it's worth benchmarking. If we fine that performance can be improved by avoiding the FFI in certain cases, I think the longer-term goal should be to get any tweaks upstreamed into GHC itself so that other packages can benefit from them. Ideally, people should just be able to use `compareByteArrays#` and trust that it performs better than anything they could write.

treeowl:
I agree that upstreaming improvements is good, but I wasn't really talking about *absolute* improvements. If the primop uses the FFI, then most likely the only absolute improvement would be to stop that in favor of something inlined, likely using AVX or some such. I agree that we should delegate regardless; I'm just saying I wouldn't be surprised to see regressions in some cases along with improvements in others.

test/main.hs
```diff
@@ -92,10 +92,7 @@ byteArrayEqProp = QC.property $ \(xs :: [Word8]) (ys :: [Word8]) ->
   (compareLengthFirst xs ys == EQ) === (byteArrayFromList xs == byteArrayFromList ys)
 
 compareLengthFirst :: [Word8] -> [Word8] -> Ordering
-compareLengthFirst xs ys = case compare (length xs) (length ys) of
-  LT -> LT
-  GT -> GT
-  EQ -> compare xs ys
+compareLengthFirst xs ys = (compare `on` length) <> compare xs ys
```
treeowl: That's not going to compile. Missing `xs` and `ys`.

treeowl: Might be able to write `compareLengthFirst = (compare `on` length) <> compare`, but that might go too far.

andrewthad: fixed

andrewthad:
I've added some benchmarks for comparison functions on [the benchmark PR](https://github.com/haskell/primitive/pull/114). These should be helpful for deciding what additional optimizations may be useful.

Data/Primitive/ByteArray.hs
andrewthad: I tried to start on this, and I realized I don't actually know how to make something that runs in `ST` a good list consumer. I know it requires `foldr`, and I'll want to delay inlining. In this context, would the specialized type of `foldr` look like this:

    foldr :: Prim a => (a -> ST s (MutableByteArray s) -> ST s (MutableByteArray s)) -> ST s (MutableByteArray s) -> [a] -> ST s (MutableByteArray s)

And then perform the freeze and the `runST` after this. Is that right?

andrewthad:
I've open [an issue](https://github.com/haskell/primitive/issues/120) to track progress on the `byteArrayFromList` question earlier. Also, I believe this is ready.

RyanGlScott:
This looks good to me upon a quick glance. Do you agree, @treeowl?

Data/Primitive/ByteArray.hs
treeowl: Let's deal with this question in a separate issue.

treeowl:
@RyanGlScott, yes, it looks good to me.

Issue #75, opened 2018-01-31
PrimUnlifted for MVar and TVar?


Zemyla:
The `PrimUnlifted` typeclass seems to be about types that are wrappers around unboxed, garbage-collected pointers. `MVar` and `TVar` seem to fit the bill. Is there a reason why instances for them couldn't be written like so:

```haskell
import GHC.MVar (MVar(..))
import GHC.Conc (TVar(..))

instance PrimUnlifted (MVar a) where
  toArrayArray# (MVar mv#) = unsafeCoerce# mv#
  fromArrayArray# mv# = MVar (unsafeCoerce# mv)

instance PrimUnlifted (TVar a) where
  toArrayArray# (TVar tv#) = unsafeCoerce# tv#
  fromArrayArray# tv# = TVar (unsafeCoerce# tv)
```

Given that there's an instance for MutVar, it doesn't seem to be limited to the types that are shown in `GHC.Prim` as being usable with `ArrayArray`s. Is it because it doesn't deal with types that aren't directly defined in the `primitive` package, like `STRef`s and `IORef`s either? `STRef` and `IORef` seem to be subsumed by `MutVar`, but `MVar` and `TVar` don't have a substitute in `primitive`.

andrewthad:
Anything of kind `TYPE 'UnliftedRep` is allowed in an `UnliftedArray`. In fact, I have a [ghc proposal to make this more clear](https://github.com/ghc-proposals/ghc-proposals/pull/112). The instances that you suggest are safe.

andrewthad:
I've put up a PR for this.

PR #119, opened 2018-04-01
add PrimUnlifted instances for TVar and MVar

andrewthad:
Resolves https://github.com/haskell/primitive/issues/75

### DIFF
--- a/Data/Primitive/UnliftedArray.hs
+++ b/Data/Primitive/UnliftedArray.hs
@@ -82,6 +82,9 @@ import qualified Data.Primitive.SmallArray as SA
 import qualified Data.Primitive.MutVar as MV
 import qualified GHC.MVar as GM (MVar(..))
 import qualified GHC.Conc as GC (TVar(..))
+import qualified GHC.Stable as GSP (StablePtr(..))
+import qualified GHC.Weak as GW (Weak(..))
+import qualified GHC.Conc.Sync as GCS (ThreadId(..))
 
 -- | Immutable arrays that efficiently store types that are simple wrappers
 -- around unlifted primitive types. The values of the unlifted type are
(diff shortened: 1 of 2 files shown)


cartazio:
This sound non controversial.  I’ll check it this afternoon

RyanGlScott:
This looks sensible to me.

I wonder: is there a comprehensive list of primitive types somewhere that have return kinds of `TYPE UnliftedRep`? It's possible that we've overlooked others, but this patch certainly seems fine in isolation.

RyanGlScott:
I think I answered my own question: to find all such `TYPE UnliftedRep`-returning types, you'd need to search http://git.haskell.org/ghc.git/blob/5ab8094e4579c08973260c2d18599be0738526ec:/compiler/prelude/TysPrim.hs#l204 and find all uses of `UnliftedRep` (since they're primitive types, their types are all hard-wired directly into the compiler).

Again, I'm not suggesting you do this in this pull request, but it might be worth a separate audit.

andrewthad:
@RyanGlScott Thanks. I went through that file and picked out three more: `Weak`, `ThreadId`, and `StablePtr`. I'm intentionally going to leave out `BCO#` because it's new and I don't know if it's expected to change any. Also, `Compact#` has to be left out because the the data constructor that wraps it takes several other fields. Sadly, `StableName#` must be omitted because nothing in `base` exports the data constructor of `StableName`.

RyanGlScott:
Awesome! Thanks for being thorough, @andrewthad.

@cartazio, does this patch look good to you?

treeowl:
Ryan, PrimUnlifted has convinced me that you were right: GHC should forbid
hand-written Generic instances. That would allow PrimUnlifted to have
generic default definitions for its methods. As it is, it can't, because
that would be unsafe. I think you should write a GHC proposal for that, or
I can.

RyanGlScott:
Silly question: is there a particular hand-written `Generic` instance you had in mind when writing that comment? (I certainly can believe that you could end up with shenanigans, but the exact nature of the shenanigans is unclear to me.)

If you're feeling up to writing a proposal, I'd definitely support it :)

treeowl:
Oh, wait a minute... I may have been wrong about that, now that I think of
it! Maybe it *can* be done safely. Let me give it a go. That won't help any
for the standard types that don't have Generic instances, but may help for
hand-written ones.

treeowl:
Ryan, do you know why we can derive Generic for types with constructors
containing Addr# but not ones containing, e.g., Array#? This seems a bit
weird!

RyanGlScott:
It's a long story. A good place to start is https://ghc.haskell.org/trac/ghc/ticket/10868.

Several stock classes, such as `Eq`, `Ord`, and `Show`, can be derived for data types with fields of certain unboxed types. These are generally of one the following six types:

* `Addr#`
* `Char#`
* `Double#`
* `Int#`
* `Float#`
* `Word#`

Because deriving `Generic` wasn't able to deal with these, we added the `URec` data family for `GHC.Generics`, and gave it six corresponding data family instances for each unboxed type.

Thus, deriving `Generic` can only handle those six particular primitive types at the moment since, well, that's all we gave it the power to do so. `URec` was made a data family so that in the future, we could extend it with the ability to work over more unboxed types if the situation called for it. (So we could conceivably add a case for `Array#`.)

That being said, I've never been super-happy with the current design. Perhaps it could be cleaned up now that we have levity polymorphism...

Issue #121, opened 2018-04-02
Missing `Prim` instances for `Foreign.C.Types`


TerrorJack:
Various types in `Foreign.C.Types` don't have `Prim` instances, it would be handy to add them. Mind me opening a PR for this?

treeowl:
Open away!

RyanGlScott:
Fixed in #123.

PR #123, opened 2018-04-02
Additional Prim instances for newtypes in base

TerrorJack:
This addresses #121

### DIFF
--- a/Data/Primitive/Types.hs
+++ b/Data/Primitive/Types.hs
@@ -208,6 +208,8 @@ derivePrim(FunPtr a, FunPtr, sIZEOF_PTR, aLIGNMENT_PTR,
            indexAddrArray#, readAddrArray#, writeAddrArray#, setAddrArray#,
            indexAddrOffAddr#, readAddrOffAddr#, writeAddrOffAddr#, setAddrOffAddr#)
 
+-- This is a workaround, since GeneralizedNewtypeDeriving doesn't work for Prim.
+-- newtypePrim(ty, ctr) works if the wrapped type has a Prim instance.
 #define newtypePrim(ty, ctr)                                    \
 instance Prim ty where {                                        \
   sizeOf# (ctr x) = sizeOf# x                                   \
@@ -234,8 +236,10 @@ instance Prim ty where {                                        \
 ; {-# INLINE setOffAddr# #-}                                    \
 }
 
+-- It's often the case that the newtype constructor has the same name as the type.
 #define newtypeHomoPrim(ty) newtypePrim(ty, ty)
 
+-- Prim instances for newtypes in Foreign.C.Types
 newtypeHomoPrim(CChar)
 newtypeHomoPrim(CSChar)
 newtypeHomoPrim(CUChar)
@@ -260,12 +264,12 @@ newtypeHomoPrim(CIntMax)
 newtypeHomoPrim(CUIntMax)
 newtypeHomoPrim(CClock)
 newtypeHomoPrim(CTime)
-#if MIN_VERSION_base(4,4,0)
 newtypeHomoPrim(CUSeconds)
 newtypeHomoPrim(CSUSeconds)
-#endif
 newtypeHomoPrim(CFloat)
 newtypeHomoPrim(CDouble)
+
+-- Prim instances for newtypes in System.Posix.Types
 newtypeHomoPrim(CDev)
 newtypeHomoPrim(CIno)
 newtypeHomoPrim(CMode)
--- a/Data/Primitive/Types.hs
+++ b/Data/Primitive/Types.hs
@@ -21,6 +21,8 @@ module Data.Primitive.Types (
 import Control.Monad.Primitive
 import Data.Primitive.MachDeps
 import Data.Primitive.Internal.Operations
+import Foreign.C.Types
+import System.Posix.Types
 
 import GHC.Base (
     Int(..), Char(..),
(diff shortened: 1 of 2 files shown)


Data/Primitive/Types.hs
```diff
@@ -205,3 +207,72 @@ derivePrim(Ptr a, Ptr, sIZEOF_PTR, aLIGNMENT_PTR,
...
+#if MIN_VERSION_base(4,10,0)
+newtypeHomoPrim(CBool)
+#endif
+newtypeHomoPrim(CIntPtr)
+newtypeHomoPrim(CUIntPtr)
+newtypeHomoPrim(CIntMax)
+newtypeHomoPrim(CUIntMax)
+newtypeHomoPrim(CClock)
+newtypeHomoPrim(CTime)
+#if MIN_VERSION_base(4,4,0)
```
RyanGlScott: `primitive` requires `base-4.5` or later, so there's no need for this bit of CPP.

Data/Primitive/Types.hs
```diff
@@ -208,6 +208,8 @@ derivePrim(FunPtr a, FunPtr, sIZEOF_PTR, aLIGNMENT_PTR,
            indexAddrArray#, readAddrArray#, writeAddrArray#, setAddrArray#,
            indexAddrOffAddr#, readAddrOffAddr#, writeAddrOffAddr#, setAddrOffAddr#)
 
+-- This is a workaround, since GeneralizedNewtypeDeriving doesn't work for Prim.
```
RyanGlScott: Why doesn't `GeneralizedNewtypeDeriving` work for `Prim`, out of curiosity?

TerrorJack: @RyanGlScott Don't know either, it was giving a compile error.

RyanGlScott: I ask because this code works for me:

```haskell
{-# LANGUAGE CPP #-}
{-# LANGUAGE GeneralizedNewtypeDeriving #-}
{-# LANGUAGE MagicHash #-}
{-# LANGUAGE UnboxedTuples #-}

#if __GLASGOW_HASKELL__ >= 800
{-# LANGUAGE TypeInType #-}
#endif
module Bug where

import GHC.Exts

class Prim a where
  readByteArray# :: MutableByteArray# s -> Int# -> State# s -> (# State# s, a #)

instance Prim Char

newtype MyChar = MyChar Char
  deriving Prim
```

So why wouldn't it work here?

RyanGlScott: Right. But that's my point—the error message tells you to enable `TypeInType`. So why not do so, as in the code in https://github.com/haskell/primitive/pull/123#discussion_r178567461?

treeowl: @RyanGlScott, we can only use `TypeInType` with new GHC versions, so I don't think that's a viable option.

TerrorJack: @RyanGlScott I'm very skeptical about introducing an extension like `TypeInType` in such a trivial use case. But seems you're not concerned so using that anyway.

RyanGlScott: I disagree. Let me give some context on this error message.

You'll only ever experience this with GHC 8.4.1 or later, since that's the first version that featured a fix for https://ghc.haskell.org/trac/ghc/ticket/14579. The content of the #14579 isn't terribly important, except for the fact that it fixed a bug in `GeneralizedNewtypeDeriving` by explicitly inserting kinds in places that might need them.

A consequence of this change is that if you're GND'ing a method that involves unboxed tuples, such as `readByteArray#`, then you'll generate code that has explicit `TYPE (TupleRep ...)` kinds. This requires the use of `TypeInType`, hence the error message.

Thus, we only need to enable `TypeInType` on newer GHC versions, since it's a problem that's unique to newer GHC versions. (Technically, we could just as well do `#if __GLASGOW_HASKELL__ >= 804`, but I picked `800` since that's the version GHC version that actually sported `TypeInType`.)

treeowl:
Ahhhh. In that case, go right ahead.

Data/Primitive/Types.hs
TerrorJack: Thanks for the clarification. GHC really is an adventure :)

Data/Primitive/Types.hs
```diff
@@ -205,3 +211,45 @@ derivePrim(Ptr a, Ptr, sIZEOF_PTR, aLIGNMENT_PTR,
...
+deriving instance Prim CDev
+deriving instance Prim CIno
+deriving instance Prim CMode
+deriving instance Prim COff
+deriving instance Prim CPid
+deriving instance Prim CSsize
+#if MIN_VERSION_base(4,10,0)
+deriving instance Prim CClockId
+#endif
+deriving instance Prim Fd
```
RyanGlScott: This surely isn't all of the newtypes found in `System.Posix.Types`, right? I'm looking at http://hackage.haskell.org/package/base-4.11.0.0/docs/System-Posix-Types.html, and there's a good deal more of them than listed here.

Also, not all of these types are exposed on all operating systems. You'll need to guard each one with CPP as shown in http://hackage.haskell.org/package/base-4.11.0.0/docs/src/System.Posix.Types.html. For instance:

```haskell
#if defined(HTYPE_DEV_T)
deriving instance Prim CDev
#endif
...
```

RyanGlScott:
Thanks, @TerrorJack!

PR #128, opened 2018-04-04
Test resizeByteArray

andrewthad:
Write tests to ensure that the implementation of resizeByteArray on older GHCs, which uses a shim, behaves the same as the implementation on newer GHCs, which uses a primop.

Resolves https://github.com/haskell/primitive/issues/115.


RyanGlScott:
Thanks, @andrewthad. Do you mind adding brief comments to each property describing what they are testing?

andrewthad:
Done.

Issue #134, opened 2018-04-19
Consistent naming for traversals in PrimMonad


andrewthad:
Now that the PrimArray branch has landed in master, we have to reconcile naming. I've added some documentation to `Data.Primitive` that explains the naming convention that I propose for traversals/filters/maps in `primitive`. These conventions suggest that `unsafeTraverseArray` and `unsafeTraverseSmallArray` should be named `traverseArrayP` and `traverseSmallArrayP` respectively instead. These haven't been released to hackage yet, so it's still easy to rename them. I wanted to gauge consensus on this. Alternatively, we could pick a different naming scheme (ie- prefix with `unsafe` instead of suffix with `P`), and I could just change the function in `PrimArray` and the docs. @treeowl I think you wrote these two functions, so I would particularly like your input.

treeowl:
I did write those functions. I like having the word `unsafe` in the name, but I'm also okay with the `P` suffix. Should we cut the baby in half and use `unsafeTraverseArrayP`?

andrewthad:
The `P` suffix comes from a desire for brevity. So, doing both would be rather analogous to cutting the baby in half.

I think my mine gripe with the word unsafe is that, when I use these functions, I don't really want to see the word unsafe all over my code. Yes, it's unsafe in the sense that doing it in `ListT IO` will make it not work correctly, but I'm almost universally going to use this in `ST s`, and so the warning for caution just wastes horizontal space on my screen. There are way more easily used unsafely (like the indexing functions) that don't get branded as unsafe.

Another very minor gripe of mine is that the `itraverse` functions start to look weird. Do we do `unsafeITraversePrimArray` or `unsafeItraversePrimArray`?

Maybe we could use unsafe as a suffix. This is uncommon, but:

- `traverseArrayUnsafely`
- `itraversePrimArrayUnsafely`

treeowl:
I guess I can learn to live with just the P suffix. Someone writing a safe
interface can always rename things as needed.

cartazio:
If the unsafe is with respect to the linearity details, all of pritmive is
unsafe.  Any backtracking is not a supported mode of operation even though
the code can support it via the generic monad structures

PR #137, opened 2018-04-19
rename unsafe traversals

andrewthad:
Resolves https://github.com/haskell/primitive/issues/134

### DIFF
--- a/Data/Primitive/Array.hs
+++ b/Data/Primitive/Array.hs
@@ -23,7 +23,7 @@ module Data.Primitive.Array (
   cloneArray, cloneMutableArray,
   sizeofArray, sizeofMutableArray,
   fromListN, fromList,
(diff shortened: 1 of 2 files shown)


treeowl:
CI fails. It looks like you missed a spot. Use git grep.

cartazio:
The comments all mention the term affine but don’t explain it anywhere.  Recall that I myself was confused by your use of affine until I realized you meant it in the linear logical sense

andrewthad:
Good point. I'll amend that in another PR.

cartazio:
Great!

Could you also clarify the listt remark?  Eg backtracking / nondeterminism
monads such as ListT or LogicT?

Though this per se is possibly true for almost any code written on top of
prim monad.  Though I agree this sorta code is a likely culprit

PR #141, opened 2018-04-20
add filterArray

andrewthad:
WIP. I still need to add this for `SmallArray` as well.

### DIFF
--- a/Data/Primitive/Array.hs
+++ b/Data/Primitive/Array.hs
@@ -24,6 +24,7 @@ module Data.Primitive.Array (
   sizeofArray, sizeofMutableArray,
   fromListN, fromList,
   mapArray',
+  filterArray,
   traverseArrayP
 ) where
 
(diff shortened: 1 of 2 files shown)


treeowl:
I recommend a different implementation. Allocate a bit vector (e.g. `MutableByteArray`; see the implementation of unboxed arrays of `Bool` in the `array` package) the same size as the original array. Use that to record which elements to keep, and keep track of the count. Then allocate a target array of exactly the right size and copy. The big advantages:

1. The intermediate array is roughly one sixty-fourth as big.
2. There's no extra mutation of *boxed* arrays, which should help keep the garbage collector happy.

andrewthad:
The implementation you describe sounds better. I'll work on that soon.

chessai:
here is the filter/ifilter from `contiguous-0.4` (unreleased):

```haskell
-- | Drop elements that do not satisfy the predicate.
filter :: (Contiguous arr, Element arr a)
  => (a -> Bool)
  -> arr a
  -> arr a
filter p arr = ifilter (\_ a -> p a) arr
{-# inline filter #-}

-- | Drop elements that do not satisfy the predicate which
--   is applied to values and their indices.
ifilter :: (Contiguous arr, Element arr a)
  => (Int -> a -> Bool)
  -> arr a
  -> arr a
ifilter p arr = runST $ do
  marr :: MutablePrimArray s Word8 <- newPrimArray sz
  let go1 :: Int -> Int -> ST s Int
      go1 !ix !numTrue = if ix < sz
        then do
          atIx <- indexM arr ix
          let !keep = p ix atIx
          let !keepTag = I# (dataToTag# keep)
          writePrimArray marr ix (fromIntegral keepTag)
          go1 (ix + 1) (numTrue + keepTag)
        else pure numTrue
  numTrue <- go1 0 0
  if numTrue == sz
    then pure arr
    else do
      marrTrues <- new numTrue
      let go2 !ixSrc !ixDst = if ixDst < numTrue
            then do
              atIxKeep <- readPrimArray marr ixSrc
              if isTrue atIxKeep
                then do
                  atIxVal <- indexM arr ixSrc
                  write marrTrues ixDst atIxVal
                  go2 (ixSrc + 1) (ixDst + 1)
                else go2 (ixSrc + 1) ixDst
            else pure ()
      go2 0 0
      unsafeFreeze marrTrues 
  where
    !sz = size arr
{-# inline ifilter #-}
```

PR #142, opened 2018-04-20
traverse Array with Maybe more quickly

andrewthad:
WIP. I still need to handle `SmallArray`. I also need to make this work with `Either`, but the backwards-compatibility story there is more annoying because of the historical absence of `ExceptT`. I also want to do this for `State`. Probably not `Writer` though since it has abysmal performance anyway.

### DIFF
--- a/Data/Primitive/Array.hs
+++ b/Data/Primitive/Array.hs
@@ -74,6 +74,8 @@ import Text.ParserCombinators.ReadP
 import Data.Functor.Classes (Eq1(..),Ord1(..),Show1(..),Read1(..))
 #endif
 
+import Control.Monad.Trans.Maybe (MaybeT(MaybeT,runMaybeT))
+
 -- | Boxed arrays
 data Array a = Array
   { array# :: Array# a }
@@ -524,6 +526,8 @@ traverseArray f = \ !ary ->
    traverseArrayP f
 "traverse/IO" forall (f :: a -> IO b). traverseArray f =
    traverseArrayP f
+"traverse/Maybe" forall (f :: a -> Maybe b). traverseArray f =
+   (\xs -> runST (runMaybeT (traverseArrayP (MaybeT . return . f) xs)))
  #-}
 #if MIN_VERSION_base(4,8,0)
 {-# RULES
@@ -532,6 +536,8 @@ traverseArray f = \ !ary ->
            -> Array a -> Identity (Array b)) (fmap f)
  #-}
 #endif
+
+
 
 -- | This is the fastest, most straightforward way to traverse
 -- an array, but it only works correctly with a sufficiently


treeowl:
I like this. For `Either`, you can either copy the basics of `ExceptT` from `transformers` or (perhaps better) just manually work it into the code for `traverseArrayP`.

Side note: it's most unfortunate for `RULES` that we don't have, e.g., `Applicative (ExceptT e m) :- Applicative m`, and that we have no way to match on the use of a particular instance dictionary either. So all these rules have to be pretty much monomorphic.

treeowl:
I don't understand why you think the manual copies (with rewrite rules to match) wouldn't help.

andrewthad:
I think GHC will produce the same core.

treeowl:
Quite possibly, but you won't be depending on having ExceptT in
transformers.

andrewthad:
Ah, it was consistent behavior with all versions of containers that was your concern. I've changed it as you suggested. Also, I added a benchmark just to see if there was a performance difference between the two implementations. They have identical performance (Wooh GHC!).

andrewthad:
> All these rewrite rules (including the ones I wrote) feel somewhat unsatisfactory, because they completely fall down for transformers. I'm wondering if there might be a way to catch known transformations of known base monads, using rewrite rules with fall-backs. 

I can't find a way to make this work. Consider what a rewrite rule for `MaybeT` might look like:

   "traverse/MaybeT" forall (f :: a -> MaybeT m b). traverseArray f = ...

Here's the problem: in the rewrite rule, we have no way to tell GHC that `m` needs to have a `PrimMonad` constraint. GHC appears to have a syntax that tricks the user into thinking they can do this, but it doesn't actually work.

andrewthad:
In fact, in the rewrite rule, GHC doesn't even know that `m` has an `Applicative` instance.

treeowl:
You certainly can't do it that way. I'm not sure if you can do it at all or
not, but if so it will require rewrite rules "all the way down" to a base
type a rule matches on.

treeowl:
The basic idea would be to step through the layers of monad transformers
you recognize, recording how each one transforms a monad. Then at the
bottom, find a base monad you recognize and build up the ultimate bind and
return. I'm not sure this can be done, but I strongly suspect it can.

andrewthad:
I'm still not able to follow how this is supposed to look. Going back to `MaybeT`:

    "traverse/MaybeT" forall (f :: a -> MaybeT m b). traverseArray f = ...

I cannot see anything that could go on the RHS (other than `traverseArray f`) that is both correct and satisfies the type checker.

andrewthad:
Wait, I think I might be beginning to see a way.

treeowl:
Wait, I think I might be beginning to see a way.
>

Me too 😉

treeowl:
The race is on.

treeowl:
I think we can get MaybeT, ExceptT, and StateT, but not WriterT, RWST, or
ErrorT (since those have side constraints). I haven't actually tested my
rules, though, so I don't know if they fire or, if they do, whether they
produce decent code.

andrewthad:
Have you pushed these somewhere?

treeowl:
No. Thus far they're pretty much faked up. When I'm home I'll (try to) make
them a bit less fake and push to show you.

treeowl:
Until then, I strongly urge you to give it a good try. You may well come up
with a better idea than I did, and in any case you'll certainly learn some
things.

cartazio:
It feels like we’re bordering on feature / ideas for improving rules based
optimization. What is it we wish we could say / can only express via hand
scripted hermit tool optimization’s?

David: have you tried out hermit?

treeowl:
@cartazio, I've only looked at hermit briefly. It's rather complicated. The basic limitation is basically that the instance resolution derivations drop away after type checking. So for example the simplifier doesn't know how GHC resolved `PrimMonad (WriterT w m)` and therefore doesn't know `Monad m` or `Monoid w`.

cartazio:
Is there ways this could be changed in ghc?

Seems like it would be at least a useful discussion as feature request
motivated by what optimization’s you want to be able to say simply.

I’d like it if we can keep primitive simple if we can.  At some point too
much rules optImitation can hinder easy of changing internals.  Though we
don’t have much internals going on :)

treeowl:
For stacks based on `ST` or `IO`, my version seems *much* simpler, and involves far fewer enormous piles of code that need to go away. I suggest you just ignore those stacks altogether and try to come up with a solution for stacks based on `Maybe`, `Either`, `Identity`, etc. It should be easy to use both solutions at the same time.

treeowl:
Separately, I'm considering a simpler general version that might work somewhat better than the one I put in before. In particular, we can go back to something somewhat list-like, but using a list that holds multiple elements per cons. I need to do some benchmarking.

andrewthad:
Undoubtedly, yours is much simpler.  In my most recent commit, I was mostly playing around just to see  what it takes to make things like Maybe and Either accepted as the base monad.  It’s helped me understand the problem better.  I  have another idea for how to extend  your solution to accept  different base monads  that I am going to try soon.

treeowl:
@cartazio, the fast way to traverse an array is to

1. Create a mutable array.
2. For each element of the given array,
    a. Perform an action to get an element and
    b. Write that element to the mutable array.
3. Freeze the mutable array and return the resulting immutable array.

We'd like `traverse` to do that whenever it can, but it's not always possible. Even ignoring safety, note that `Compose m n` is not generally a `PrimMonad`, or even a `Monad`, even if `m` and `n` both are.

cartazio:
I'm not terribly familiar with Compose, could you walk me through an example?

cartazio:
also where does Compose live?

cartazio:
hrmm

if this is the correct defintion of compose

```
-- | Right-to-left composition of functors.
-- The composition of applicative functors is always applicative,
-- but the composition of monads is not always a monad.
newtype Compose f g a = Compose { getCompose :: f (g a) }

```
wouldn't  just require PrimMonad just one of these?

cartazio:
hrmmm, that makes me think that we can't have an instance for compose, because there should only be one primonad in the stack ... i think?

cartazio:
(something something global approximate  linearity of state token)

cartazio:
i guess i'm perhaps not understnding why Compose the data type/monad enters into this, but i'll have to look at more details

treeowl:
Not even compositions of pure monads are necessarily monads. As my PR
shows, we can do good things for a lot of transformer stacks built on ST
and IO. Andrew has been working on an idea for stacks built on Identity,
Either, and Maybe  (and perhaps even (a,)), but it's not ready yet. Compose
is a generally useful way to construct Applicatives, which people certainly
use with traverse, but I don't know if we'll be able to optimize for it or
not.

treeowl:
The composition law for Traversable does seem to suggest we can probably do
something smart about Compose, actually. We know that

traverse
<http://hackage.haskell.org/package/base-4.11.1.0/docs/Data-Traversable.html#v:traverse>
(Compose
. fmap
<http://hackage.haskell.org/package/base-4.11.1.0/docs/Data-Functor.html#v:fmap>
g
. f) = Compose . fmap
<http://hackage.haskell.org/package/base-4.11.1.0/docs/Data-Functor.html#v:fmap>
 (traverse
<http://hackage.haskell.org/package/base-4.11.1.0/docs/Data-Traversable.html#v:traverse>
g)
. traverse
<http://hackage.haskell.org/package/base-4.11.1.0/docs/Data-Traversable.html#v:traverse>
 f

In particular,

traverse (Compose . f) = Compose . fmap sequenceA . traverse f

So we can turn one traversal into two, which sounds bad but is probably
actually good. For *pure* Applicatives, I conjecture that we can also turn
two traversals into one efficient one. Anyway, that's the start of a
thought in that direction. Still need to think it through a bit more.

PR #144, opened 2018-04-22
First draft of filterArray

treeowl:


treeowl:
@andrewthad, here's a draft. As you can see from my comment on `fill`, I think there's some room for improvement.

Data/Primitive/Internal/Bit.hs
```diff
@@ -0,0 +1,63 @@
...
+import Data.Bits
+
+newtype MutableBitArray s = MBA (MutableByteArray s)
+
+newBitArray :: PrimMonad m => Int -> m (MutableBitArray (PrimState m))
+newBitArray n = do
+  let s = ((n + wordSize - 1) `unsafeShiftR` 3)
+  mary <- newByteArray s
+  fillByteArray mary 0 s 0
+  return (MBA mary)
```
treeowl: Alternatively, we could refrain from filling the array here, and just use a `writeBitArray` function that takes the value to write. That's probably better, actually.

Data/Primitive/Array.hs
```diff
@@ -591,6 +595,49 @@ arrayFromListN n l =
...
+          mary <- newArray count (die "filterArray" "invalid")
+          fill 0 0 ba mary
+
+    -- This performs a few bit operations and a conditional
+    -- jump for every element of the original array. This is
+    -- not so great if most element are filtered out. We should
+    -- consider going word by word through the bit array and
+    -- using countTrailingZeroes. We could even choose
+    -- a different strategy for each word depending on its
+    -- popCount.
```
treeowl: Also: I wouldn't be surprised if `unordered-containers` had some code or ideas we could steal here.

andrewthad:
I still need to look at this. Sorry, been focused on other things the last few days. I'll try to look this over better (and add some benchmarks) later this week.

Data/Primitive/Array.hs
andrewthad: I agree that CTZ (or equivalently CLZ) is going to do best here. I'm suspect that changing strategies with `popCount` would be helpful because I cannot think of a strategy that performs better when most of the elements are preserved. However, when *all* of them corresponding to a word of bits are preserved (meaning: the word is equal to `maxBound :: Word`), we could use the functions that copy a slice of the array instead. I guess we could also do this when `popCount` is really high instead of just when it's 64 (or 32 depending on platform).

Data/Primitive/Array.hs
```diff
@@ -591,6 +595,49 @@ arrayFromListN n l =
...
+    check i count ba
+      | i /= s
+      = do
+          v <- indexArrayM arr i
+          if f v
+            then setBitArray ba i >> check (i + 1) (count + 1) ba
+            else check (i + 1) count ba
+      | otherwise
+      = do
+          mary <- newArray count (die "filterArray" "invalid")
```
andrewthad: If `count` equals the size of the original array, we have a much better option available to us. Reuse the original array. This prevents use `runArray` though.

treeowl: This is why we want `runArrays` and the like.

PR #145, opened 2018-04-23
Add indexed maps and traversals

treeowl:
* Add indexed maps (lazy and strict) and traversals (general and
  `PrimMonad`-specialized) for `Array` and `SmallArray` to match
  the ones in `PrimArray`.

* Add some missing `@since` annotations.

### DIFF
--- a/Data/Primitive/Array.hs
+++ b/Data/Primitive/Array.hs
@@ -23,8 +23,8 @@ module Data.Primitive.Array (
   cloneArray, cloneMutableArray,
   sizeofArray, sizeofMutableArray,
   fromListN, fromList,
-  mapArray',
-  traverseArrayP
+  imapArray, mapArray', imapArray',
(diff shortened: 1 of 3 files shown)


treeowl:
@andrewthad, do these look okay to you?

andrewthad:
It could use some tests like the [test for `mapArray'`](https://github.com/haskell/primitive/blob/2ad71289bb2abd2ba1af625e5a3f3a8930ceb36f/test/main.hs#L57). One of the property tests you'll want is named [imapMProp](http://hackage.haskell.org/package/quickcheck-classes-0.4.9/docs/Test-QuickCheck-Classes-IsList.html#v:imapMProp) instead of `itraverseProp`.

The `since` annotations seem too high. Should they not be `0.6.4.0` instead of `0.6.4.1`?

cartazio:
lets circle back on theses and a few others over the course of march/april (i'm round robin catching up on everything)

Issue #148, opened 2018-04-30
add MVar support to Primitive


cartazio:
this was discussed/planned years ago, but still to do

Are there other ones we should consider too?  Can we safely add StableNames or WeakRefs or Finalizers too?

cartazio:
cc @hvr @bgamari @RyanGlScott @jberryman for ideas :)

cartazio:
STM is definitely off the table because GHC's semantics doesn't support nested transactions, which would be needed to even touch the question of how to sanely consider something like  IO > STM > ST > STM in ST via primitive style PrimMonad stuff

cartazio:
Current plan for Mvar would not include forkIO. ANY efforts for
deterministic concurrency should be in a higher level library.  Unless you
think we can do that sanely and simply ... :)

Other points sound good as starting points

andrewthad:
Mildly relevant to this discussion: https://ghc.haskell.org/trac/ghc/ticket/14195

andrewthad:
I've changed my mind on `StableName`. I think generalizing it is bad for two reasons:

- Practical reason: not having access to `hashStableName` makes it nearly useless.
- Non-determinism reason: `makeStableName` is not deterministic in the presence of optimizations. Additional inlining can cause a computation that would have resulting in False to result in True.

For these reason, I don't think `StableName` is appropriate for `primitive`. Everything provided here should be deterministic.

andrewthad:
This was implemented. Closing.

PR #149, opened 2018-05-01
add a StableName that works in any PrimMonad

andrewthad:
Part of the discussion in https://github.com/haskell/primitive/issues/148.


cartazio:
The module level haddocks look Wrong ;)

Let’s put its name in the hassles
plus both of ours or certainly yours in copyright?

Or we can elide the copyright altogether ?

PR #150, opened 2018-05-01
implement MVar in PrimMonad

andrewthad:


Data/Primitive/MVar.hs
```diff
@@ -0,0 +1,117 @@
...
+-- 'readMVar' is guaranteed to receive the next 'putMVar'.
+readMVar :: PrimMonad m => MVar (PrimState m) a -> m a
+readMVar (MVar mvar#) = primitive $ \ s# -> readMVar# mvar# s#
+
+-- |Put a value into an 'MVar'.  If the 'MVar' is currently full,
+-- 'putMVar' will wait until it becomes empty.
+putMVar :: PrimMonad m => MVar (PrimState m) a -> a -> m ()
+putMVar (MVar mvar#) x = primitive $ \ s# ->
+  case putMVar# mvar# x s# of
+    s2# -> (# s2#, () #)
```
treeowl: You can just write `(# putMVar# mvar# x s#, () #)`.

andrewthad: Nice catch!

Data/Primitive/MVar.hs
```diff
@@ -0,0 +1,133 @@
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE MagicHash #-}
+{-# LANGUAGE UnboxedTuples #-}
+
+-- |
+-- Module      : Data.Primitive.ByteArray
+-- Copyright   : (c) Roman Leshchinskiy 2009-2012
```
cartazio: fix ther module name and you can drop the copyright and maintainer fields

set license for the module to BSD2 or something, style is ambiguous :) 

(we should add an authors file though?)

Data/Primitive/MVar.hs
```diff
@@ -0,0 +1,133 @@
...
+-- Portability : non-portable
+--
+-- Primitive operations on @MVar@. This module provides a similar interface
+-- to "Control.Concurrent.MVar". However, the functions are generalized to
+-- work in any 'PrimMonad' instead of only working in 'IO'. Note that all
+-- of the functions here are completely deterministic. It is only the
+-- introduction of @forkIO@ that brings nondeterminism into the picture.
+-- Users who are interesting in using these functions in 'ST' will likely
+-- want to roll their own 'ST' variant of @forkIO@. Since such a function
+-- eschews that determinism expected by users of 'ST' (and users of
```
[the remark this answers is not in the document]
cartazio: just say  "`MVar s`  users are responsible for providing abstractions or design that guarantees determinism in the presence of multi-threading"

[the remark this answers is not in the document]
cartazio: or somethign like that

Data/Primitive/MVar.hs
```diff
@@ -0,0 +1,133 @@
...
+#if __GLASGOW_HASKELL__ >= 708
+import GHC.Exts (readMVar#,tryReadMVar#)
+#endif
+
+data MVar s a = MVar (MVar# s a)
+
+instance Eq (MVar s a) where
+  (MVar mvar1#) == (MVar mvar2#) = isTrue# (sameMVar# mvar1# mvar2#)
+
+-- | Create an 'MVar' which is initially empty.
```
cartazio: create a fresh mvar in the empty state.

"an" is for things that start with a vowel

Data/Primitive/MVar.hs
```diff
@@ -0,0 +1,133 @@
...
+newEmptyMVar :: PrimMonad m => m (MVar (PrimState m) a)
+newEmptyMVar = primitive $ \ s# ->
+  case newMVar# s# of
+    (# s2#, svar# #) -> (# s2#, MVar svar# #)
+
+
+-- |Create an 'MVar' which contains the supplied value.
```
cartazio: create a new Mvar value which holds the supplied argument ?

Data/Primitive/MVar.hs
```diff
@@ -0,0 +1,133 @@
...
+#if __GLASGOW_HASKELL__ >= 708
+-- | Atomically read the contents of an 'MVar'.  If the 'MVar' is
+-- currently empty, 'readMVar' will wait until it is full.
+-- 'readMVar' is guaranteed to receive the next 'putMVar'.
+--
+-- This function is only available when compiling with GHC 7.8
+-- or newer.
+readMVar :: PrimMonad m => MVar (PrimState m) a -> m a
+readMVar (MVar mvar#) = primitive $ \ s# -> readMVar# mvar# s#
+#endif
```
cartazio: prior to 7.8 read mvar was encoded in terms of take then put, granted prior to 7.8 it would block other reads, which wasn't ideal...

whats the oldest ghc we currently support?

andrewthad: We currently support back to 7.4.

andrewthad:
I've tweaked the docs. This should address most of the concerns.

Data/Primitive/MVar.hs
andrewthad: Dang. I just realized that it's actually just not possible to shim this in a way that has the same behavior. I'm going to add a note on the main thread about this.

andrewthad:
I was contemplating a backwards-compat shim for `readMVar`, and then I found this in the docs for `readMVar` in `Control.Concurrent.MVar`:

> `readMVar` is multiple-wakeup, so when multiple readers are blocked on an `MVar`, all of them are woken up at the same time.
> *Compatibility note*: Prior to base 4.7, `readMVar` was a combination of `takeMVar` and `putMVar`. This mean that in the presence of other threads attempting to `putMVar`, `readMVar` could block. Furthermore, `readMVar` would not receive the next `putMVar` if there was already a pending thread blocked on `takeMVar`. The old behavior can be recovered by implementing `readMVar` as follows ...

The backwards-compatibility shim simply does not have the same behavior in a concurrent setting, so I think it is best to not provide `readMVar` on older GHCs.

cartazio:
the alternative is to write some CMM that messes with mvars

i'd rather avoid conditional apis that create weird gaps in client code
support

Data/Primitive/MVar.hs
treeowl: Another option is to use a different representation. When compiling with old GHC, we can do this:

```haskell
data Use = Reading | Taking

data MVar s a = MVar
  { var :: !(STRef s (Maybe a))
  , claimed :: MVar# s Use }
```

Horrible, right? Now `takeMVar` will stake a claim by *putting* `Taking` into the `MVar#`, and then set the `STRef` to `Nothing`. `putMVar` will *take* the `MVar#`. If it's `Taking`, it will then set the `STRef`. If it's `Reading`, it will instead put it back and `yield`. `readMVar` will put `Reading` into the `MVar#`, read the `STRef`, and then take the `MVar#` again.

I don't really think this is a good idea, but I figured I'd mention it.

Data/Primitive/MVar.hs
```diff
@@ -0,0 +1,133 @@
...
+-- returns immediately, with 'Nothing' if the 'MVar' was empty, or
+-- @'Just' a@ if the 'MVar' was full with contents @a@.
+--
+-- This function is only available when compiling with GHC 7.8
+-- or newer.
+tryReadMVar :: PrimMonad m => MVar (PrimState m) a -> m (Maybe a)
+tryReadMVar (MVar m) = primitive $ \ s ->
+    case tryReadMVar# m s of
+        (# s', 0#, _ #) -> (# s', Nothing #)      -- MVar is empty
+        (# s', _,  a #) -> (# s', Just a  #)      -- MVar is full
```
treeowl: You can also write a lousy shim here that tries to take and (if that succeeds) immediately puts. Not entirely non-blocking, but should work in *some* applications...

andrewthad:
I've added support for `readMVar` and `tryReadMVar` on older GHCs with a warning about the different behavior.

Data/Primitive/MVar.hs
andrewthad: Added.

Data/Primitive/MVar.hs
```diff
@@ -0,0 +1,140 @@
...
+
+-- | Atomically read the contents of an 'MVar'.  If the 'MVar' is
+-- currently empty, 'readMVar' will wait until it is full.
+-- 'readMVar' is guaranteed to receive the next 'putMVar'.
+--
+-- /Compatibility note:/ On GHCs prior to 7.8, 'readMVar' is a combination
+-- of 'takeMVar' and 'putMVar'.  This means that in the presence of
+-- other threads attempting to 'putMVar', 'readMVar' may block.
+-- Furthermore, 'readMVar' might not receive the next 'putMVar' if there
+-- is already a pending thread blocked on 'takeMVar'.
```
treeowl: What do you mean by "receive the next ..."? I think I understand the situation, but the wording is too confusing. `readMVar` may block regardless, so cut that.

Data/Primitive/MVar.hs
```diff
@@ -0,0 +1,140 @@
...
+        (# s, 0# #) -> (# s, False #)
+        (# s, _  #) -> (# s, True #)
+
+-- | A non-blocking version of 'readMVar'.  The 'tryReadMVar' function
+-- returns immediately, with 'Nothing' if the 'MVar' was empty, or
+-- @'Just' a@ if the 'MVar' was full with contents @a@.
+--
+-- This behaves differently in a concurrent setting when
+-- compiling when GHC older than 7.8. See the documentation for 'readMVar'.
+-- or newer.
```
treeowl: This is where you need the "may block" remark.

cartazio:
I’ll need to look at this all again. But why don’t we just cite / repeat
the language used in base regarding readMvar ?

treeowl:
Because the language in base is confusing. We should figure out the right thing to say, say it here, and change base to say the same.

Data/Primitive/MVar.hs
andrewthad: Unfortunately, the way that is phrased was taken from base.  I’m all right with changing it to make it more clear, but  I think it’s important to try to communicate more than just  that this function may block. It’s also not multiple-wakeup on older GHCs,  which could change the behavior of some programs in a concurrent setting.  I think that people  should refrain from writing programs that rely on this difference in behavior ( so that the emergent behavior of the system is the same either way), but it’s a difference that matters.

andrewthad: Actually, never mind, I can’t actually think of an example where you can ever get a different behavior.

treeowl: The multiple-wakeup probably isn't that important semantically, but the rest certainly is. The old-school version can actually change the value in the `MVar`.

andrewthad:
@treeowl See my revisions to the docs.

Data/Primitive/MVar.hs
```diff
@@ -0,0 +1,145 @@
...
+
+#if __GLASGOW_HASKELL__ >= 708
+import GHC.Exts (readMVar#,tryReadMVar#)
+#endif
+
+data MVar s a = MVar (MVar# s a)
+
+instance Eq (MVar s a) where
+  (MVar mvar1#) == (MVar mvar2#) = isTrue# (sameMVar# mvar1# mvar2#)
+
```
treeowl: Drop the extra parents in the patterns.

Data/Primitive/MVar.hs
```diff
@@ -0,0 +1,145 @@
...
+-- 'readMVar' is guaranteed to receive the next 'putMVar'.
+--
+-- /Multiple Wakeup:/ 'readMVar' is multiple-wakeup, so when multiple readers
+-- are blocked on an 'MVar', all of them are woken up at the same time.
+--
+-- /Compatibility note:/ On GHCs prior to 7.8, 'readMVar' is a combination
+-- of 'takeMVar' and 'putMVar'. Consequently, its behavior differs in the
+-- following ways:
+--
+-- * It is single-wakeup instead multiple-wakeup.
```
treeowl: Instead *of*

Data/Primitive/MVar.hs
```diff
@@ -0,0 +1,145 @@
...
+-- * It might not receive the value from the next call to 'putMVar' if
+--   there is already a pending thread blocked on 'takeMVar'.
```
treeowl: There's a third difference: if someone else puts a value in the `MVar` in between the take and put, then we'll overwrite that new value. Whereas the new `readMVar` never changes the `MVar`, the old one can!

Data/Primitive/MVar.hs
```diff
@@ -0,0 +1,145 @@
...
+        (# s, 0# #) -> (# s, False #)
+        (# s, _  #) -> (# s, True #)
+
+-- | A non-blocking version of 'readMVar'.  The 'tryReadMVar' function
+-- returns immediately, with 'Nothing' if the 'MVar' was empty, or
+-- @'Just' a@ if the 'MVar' was full with contents @a@.
+--
+-- /Compatibility note:/ On GHCs prior to 7.8, 'tryReadMVar' is a combination
+-- of 'tryTakeMVar' and 'putMVar'. This means that in the presence of other
+-- threads attempting to 'putMVar', 'tryReadMVar' may block.
```
treeowl: Yes, but also refer to the rest of the caveats for `readMVar`.

Data/Primitive/MVar.hs
andrewthad: If someone puts a value in the MVar in between the take and the put, then the put will block until than value is taken by some other thread. The term “overwrite” is too strong here. Similarly, I’m not yet convinced that the old definition of readMVar can change the value in an MVar. It seems like it might be able to, but I haven’t been able to come up with an example that actually demonstrates this behavior.

andrewthad:
@treeowl I've made more changes to the docs that should address the remaining concerns.

andrewthad:
Are there any other improvements to the docs I need to make?

Data/Primitive/MVar.hs
treeowl: You may not have considered enough threads. Suppose we create an empty `MVar`, then fork twice. Thread A puts A, then takes. Thread B puts B, then takes. Thread C reads. With the new `readMVar`, thread A is guaranteed to take A and thread B is guaranteed to take B. With the old one, it's a race. One possibility is that A puts, then C takes, then B puts, then A takes, then C puts, then B takes. This means A will take B and B will take A. I guess "overwrite" is the wrong word, but the read has introduced entirely new possibilities.

treeowl:
It's buried as "outdated", but I've added a comment explaining how the two `readMVar` implementations can lead to rather different results.

andrewthad:
Ah, thanks.  That makes total sense. I had only been considering examples with two threads.

andrewthad:
Are there any other requested changes, or is this good to go?

PR #152, opened 2018-05-02
add PrimMach typeclass and implement casPrimArray with it

andrewthad:
This solution to wrapping the atomics primops is an alternative to https://github.com/haskell/primitive/pull/151 and is incompatible with it. This PR only wrap `casIntArray#`, but there are a number of other functions where this typeclass could be used:

- `atomicReadIntArray#`
- `atomicWriteIntArray#`
- `fetchAddIntArray#`
- `fetchSubIntArray#`
- `fetchAndIntArray#`
- `fetchNandIntArray#`
- `fetchOrIntArray#`
- `fetchXorIntArray#`

What this PR offers is a way to write a single `casPrimArray` function such that it will work on both `PrimArray Int` and `PrimArray Word` as well as any newtype wrapper over either of these two types. (It's worth noting that the `And`, `Nand`, `Or`, and `Xor` fetch functions would typically be used with `Word`). The drawback is that a new typeclass is need to accomplish this. I would appreciate any thoughts on this.

### DIFF
--- a/Data/Primitive/PrimArray.hs
+++ b/Data/Primitive/PrimArray.hs
@@ -54,6 +54,8 @@ module Data.Primitive.PrimArray
   , getSizeofMutablePrimArray
   , sizeofMutablePrimArray
   , sizeofPrimArray
+    -- * Atomics
+  , casPrimArray
     -- * Folding
   , foldrPrimArray
   , foldrPrimArray'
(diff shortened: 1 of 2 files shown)


cartazio:
hrmm, we should circle back on these and other extant prs in a post no unlifted types primitive right?

andrewthad:
These primitives are now in http://hackage.haskell.org/package/primitive-atomic  i’m not worried about getting them in this library anymore

cartazio:
Ok great!

Should some of the stuff in Ryan newtons tool kit and some of David’s
atomic modify tools live there too? May be tricky idk

Let’s you me and chessai do a brain storming call sometime on this stuff :)

PR #153, opened 2018-05-03
Export emptyPrimArray

andrewthad:
The need for this shows up annoyingly often. We can already get an empty prim array with `mempty`, but `mempty` causes us to needlessly incur a `Monoid` constraint.


treeowl:
What `Monoid` constraint?

andrewthad:
Ah, I double lost my mind. I meant that `mempty` causes us to needlessly incur a `Prim` constraint, not a `Monoid` constraint.

Issue #155, opened 2018-05-04
mconcat broken on SmallArray


andrewthad:
I recently pushed a new release of `quickcheck-classes`, which is used in the test suite for primitive. In this release, I added additional tests for the methods `mconcat`, `sconcat`, and `stimes`, which belong to `Monoid` and `Semigroup`. I discovered that the implementation of `mconcat` for `SmallArray` is incorrect. I am going to fix this shortly.

cartazio:
Excellent !

PR #157, opened 2018-05-04
fix broken mconcat for SmallArray

andrewthad:
Resolves https://github.com/haskell/primitive/issues/155


andrewthad:
Also, I misnamed this branch.

treeowl:
Good catch.

PR #158, opened 2018-05-05
Add function for copying from Ptr to MutablePrimArray

andrewthad:
I started compiling my stuff start currently uses `prim-array` against the master branch of `primitive`, and I realized that I had forgotten to implement `copyPtrToMutablePrimArray` in `Data.Primitive.Ptr`. This happened because I was working on the `PrimArray` module and the `Ptr` module in separate branches at the same time. Anyway, this PR provides the missing wrapper around `copyAddrToByteArray#`.

### DIFF
--- a/Data/Primitive/Ptr.hs
+++ b/Data/Primitive/Ptr.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE MagicHash #-}
 {-# LANGUAGE UnboxedTuples #-}
 {-# LANGUAGE ScopedTypeVariables #-}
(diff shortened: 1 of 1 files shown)


Data/Primitive/Ptr.hs
```diff
@@ -25,10 +26,15 @@ module Data.Primitive.Ptr (
...
   copyPtr, movePtr, setPtr
+
+#if __GLASGOW_HASKELL__ >= 708
+  , copyPtrToMutablePrimArray
+#endif
 ) where
 
 import Control.Monad.Primitive
 import Data.Primitive.Types
+import Data.Primitive.PrimArray (MutablePrimArray(..))
```
RyanGlScott: This should be guarded with `#if __GLASGOW_HASKELL__ >= 708`, right?

andrewthad: I'll fix this. I've never been totally clear on the policy around warnings though. Is it something like "avoid any warnings on any version of GHC"?

RyanGlScott: > Is it something like "avoid any warnings on any version of GHC"?

Ideally that's be the case, yes. I say "ideally" since `primitive` doesn't currently meet this standard—it emits several warnings on older versions of GHC. If we really wanted to uphold this, we should enable `-Werror` and weed out the existing warnings.

That being said, this particular warning has a very simple fix, so I thought it prudent to catch it while I still could.

cartazio:
All warnings ever is tricky. But definitely any useful warnings

PR #161, opened 2018-05-23
Document the lack of guarantees around Ord instances

andrewthad:
This is another stab at what https://github.com/haskell/primitive/pull/68 tried to do. I've gone a little further though. I've added similar warnings around *all* of the `Ord` instances. This is important because in a future version of `primitive`, we should probably change the `Ord` instance for `Array` to compare the lengths first.

### DIFF
--- a/Data/Primitive/Array.hs
+++ b/Data/Primitive/Array.hs
@@ -379,6 +379,9 @@ arrayLiftCompare elemCompare a1 a2 = loop 0
     = elemCompare x1 x2 `mappend` loop (i+1)
     | otherwise = compare (sizeofArray a1) (sizeofArray a2)
 
+-- | __Note__: This library provides no guarantees about the particular ordering
+-- used. Users should not rely on it being the same as the ordering used for
+-- lists. It is subject to change between major versions.
 instance Ord a => Ord (Array a) where
   compare a1 a2 = arrayLiftCompare compare a1 a2
 
--- a/Data/Primitive/ByteArray.hs
+++ b/Data/Primitive/ByteArray.hs
@@ -446,12 +446,9 @@ instance Eq ByteArray where
       n1 = sizeofByteArray ba1
       n2 = sizeofByteArray ba2
 
--- Note: On GHC 8.4, the primop compareByteArrays# performs a check for pointer
--- equality as a shortcut, so the check here is actually redundant. However, it
--- is included here because it is likely better to check for pointer equality
--- before checking for length equality. Getting the length requires deferencing
--- the pointers, which could cause accesses to memory that is not in the cache.
--- By contrast, a pointer equality check is always extremely cheap.
+-- | __Note__: This library provides no guarantees about the particular ordering
+-- used (lexicographic,alphabetical,etc.). Users should not rely on the ordering,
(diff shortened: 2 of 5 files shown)


treeowl:
Why do you think we should change those instances? I really disagree. I
think people who just want a meaningless instance to store things in Maps
should use newtypes.

andrewthad:
If we compare the lengths first, we can make the instances faster. My viewpoint is that the `Ord` instances in `primitive` should be chosen based on whatever performs best, not by compatibility with the instances for lists. Already, the library communicates a sort of inconsistent on this. `Array` and `SmallArray` use the same ordering that list does, but `ByteArray` eschews compatibility with the `Ord` instance for `[Word8]` in favor of performance. The unreleased instance for `UnliftedArray` follows the `Array`-style `Ord` instances, while `PrimArray` follows `ByteArray`'s example.

treeowl:
I believe we should make the instances for the bare types match those for
lists, old-school arrays, vectors, sequences, and everything else under the
sun. I think it would also make good sense to offer a new type for each to
do the faster thing.

cartazio:
Hrmm.  I’ll carve some time to scribble about this and make a choice for
interim.

And the end of the day primitive is well primitive. And choices should be
made that don’t break vector and similar high profile down stream libs.
Subject to that constraint everything should be as fast and simple cost
model wise as possible.

Eg if we can use unsafe ptr equality to accelerate our Rw and ord fast
paths great.

andrewthad:
David mentioned that basically all other sequence data structure using the same ordering. I find this moderately compelling. I think that, at least for the pending release, `PrimArray` should be changed to use the same ordering as `Array`, `SmallArray`, etc. However, `ByteArray` should be left alone since messing with it would be a breaking change.

cartazio:
1) i'm looking at the code right now, and perhaps i'm overlooking something, but in current master what examples would Ord for Array vs PrimArray differ wrt elements etc?

2) independent of the changes or not to the code,  lets also document what the Ord instance IS for each of these, they're subject to change, but lets make them intelligible per major version

cartazio:
(i'm juggling some code / design bits unrelated to this repo atm so my ability to see what the differences are isn't 100% :) )

andrewthad:
Currently, array has a lexicographic ordering. PrimArray’s implementation of compare checks the length of each  argument  and uses this information to short-circuit.

    [7,9]
    [3,4,5]

 Under PrimArray’s  current notion of ordering, the second array is bigger since its length is bigger.  Array’s  ordering would deem the first one bigger cents a comparison of the first element of each array shows it  having a bigger first element.

 Whatever we do, we should probably try to do it as consistently as possible, so I think PrimArray  should be changed to match the behavior of the others ( except for ByteArray,  which could perhaps be changed in the future  with a major version bump )

cartazio:
Vs the order induced by list of a , aka lexicographic order

cartazio:
Let’s document what each of these current instances does explicitly.  Plus
warn that we don’t guarantee the ordering Choice is stable between major
versions

andrewthad:
This has been superceeded by https://github.com/haskell/primitive/pull/162

PR #162, opened 2018-05-24
Ord Instance Documentation

andrewthad:
This superceeds https://github.com/haskell/primitive/pull/161 It makes `PrimArray`'s `Ord` instance use a lexicographic ordering, and it documents what ordering all of the array types use for their `Ord` instances.


RyanGlScott:
This sounds good to me. Do you agree, @treeowl?

PR #164, opened 2018-05-24
Disable semirings support in quickcheck-classes

andrewthad:
@RyanGlScott Can you look over this? @chessai recently added support in `quickcheck-classes` for checking instances of `Semiring`. There's a flag to disable it, and I just want to make sure that I've modified all the appropriate places in `primitive` to stop travis from building `semirings`.

### DIFF
--- a/test/primitive-tests.cabal
+++ b/test/primitive-tests.cabal
@@ -36,7 +36,7 @@ test-suite test
                , tasty-quickcheck
                , tagged
                , transformers >= 0.3
-               , quickcheck-classes >= 0.4.11
+               , quickcheck-classes >= 0.4.11.1
   ghc-options: -O2
 
 source-repository head


RyanGlScott:
Oh, right. To make this work, you'll have to regenerate the `.travis.yml` script using `hakell-ci` (with the command it puts at the top).

However, I recall having to make manual tweaks after it was generated to satisfy `primitive`'s needs. Let me see if I can reduce the amount of manual intervention going forward...

RyanGlScott:
I tried my hardest, but I wasn't able to automate [these lines](https://github.com/haskell/primitive/pull/164/files#diff-354f30a63fb0907d4ad57269548329e3R97). This is the subject of https://github.com/haskell-CI/haskell-ci/issues/129, which @phadej has offered to write a patch for, so I think I'll just wait for that to land before I try this again.

That being said, I took a look at your patch, and I believe that everything was changed correctly. Let's see if Travis agrees, and if so, feel free to merge.

RyanGlScott:
It appears that `quickcheck-classes` isn't building on older versions of GHC now.

PR #166, opened 2018-05-24
add since metadata from 0.6 to 0.6.3

chessai:
@andrewthad added `since` metadata for everything coming up in 0.6.4.0, this PR adds stuff for 0.6.0.0 to 0.6.3.0

### DIFF
--- a/Control/Monad/Primitive.hs
+++ b/Control/Monad/Primitive.hs
@@ -78,6 +78,8 @@ class Monad m => PrimMonad m where
 -- Unlike 'PrimMonad', this typeclass requires that the @Monad@ be fully
 -- expressed as a state transformer, therefore disallowing other monad
 -- transformers on top of the base @IO@ or @ST@.
+--
+-- @since 0.6.0.0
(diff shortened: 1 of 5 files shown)


andrewthad:
The annotations for `sizeOf` and `alignment` are a little misleading. These functions have [existed since `0.1`](http://hackage.haskell.org/package/primitive-0.1/docs/Data-Primitive.html), but they were defined in `Data.Primitive`, and then they got moved into `Data.Primitive.Types` in `0.6.3`. Perhaps it would be best to just explicitly state that in version of `primitive` prior to `0.6.3`, these function were only available through `Data.Primitive`.

chessai:
@andrewthad I just pushed the discussed change

andrewthad:
This looks good to me. I'll wait a day for anyone else to weigh in and then merge if there are no concerns voiced.

andrewthad:
The build failure is just the result of an issue with `quickcheck-classes` that has since been fixed. Merging.

PR #181, opened 2018-07-05
ban vim swap files in .gitignore

chessai:


RyanGlScott:
I'm generally of the opinion that these sorts of things should be placed in a global `.gitignore` file, as they're rather language-agnostic.

chessai:
That makes sense. I think I mostly agree with that.

andrewthad:
I agree with David. I don't mind having them here even though people should have these globally.

chessai:
Can this just be merged then? There are no breaking changes presented, just 3 lines in a .gitignore

PR #184, opened 2018-07-06
add @since pragmas for newtypes

chessai:
I forgot to add since pragmas for the newtype instances that were just merged. This PR adds those


andrewthad:
The next version cannot be 0.6.4.1 as this PR suggests. The fourth component of a version is reserved for changes that do not affect the API at all, like bug fixes. Since typeclass instances have been added, the next version must be 0.6.5.0 (although it very well may be 0.7.0.0).

chessai:
Is this OK to be merged? I fixed the inaccuracy

PR #186, opened 2018-07-15
foldMap, foldMap' for PrimArray

chessai:


andrewthad:
I'm good with this. If there aren't any objection in the next few days, I'll merge this.

RyanGlScott:
No objections here, although these deserve to be mentioned in the changelog.

chessai:
Sorry I keep forgetting that. I'll add it in.

chessai:
Is this good to be merged?

Data/Primitive/PrimArray.hs
```diff
@@ -467,6 +469,16 @@ sizeofPrimArray :: forall a. Prim a => PrimArray a -> Int
 {-# INLINE sizeofPrimArray #-}
 sizeofPrimArray (PrimArray arr#) = I# (quotInt# (sizeofByteArray# arr#) (sizeOf# (undefined :: a)))
 
+-- | Lazily map each element of the primitive array to a monoid, and combine the results.
+foldMapPrimArray :: forall a m. (Prim a, Monoid m) => (a -> m) -> PrimArray a -> m
+{-# INLINE foldMapPrimArray #-}
+foldMapPrimArray f = foldrPrimArray (\a acc -> acc `mappend` f a) mempty
+
+-- | Strictly map each element of the primitive array to a monoid, and combine the results.
+foldMapPrimArray' :: forall a m. (Prim a, Monoid m) => (a -> m) -> PrimArray a -> m
```
treeowl: The documentation is insufficient. It should explain which way the combination associates. Also, it's notably not (necessarily) strict in the function results. The documentation should reflect this and there should be a comment explaining that decision.

Data/Primitive/PrimArray.hs
```diff
@@ -467,6 +469,16 @@ sizeofPrimArray :: forall a. Prim a => PrimArray a -> Int
 {-# INLINE sizeofPrimArray #-}
 sizeofPrimArray (PrimArray arr#) = I# (quotInt# (sizeofByteArray# arr#) (sizeOf# (undefined :: a)))
 
+-- | Lazily map each element of the primitive array to a monoid, and combine the results.
+foldMapPrimArray :: forall a m. (Prim a, Monoid m) => (a -> m) -> PrimArray a -> m
```
treeowl: The documentation is insufficient. It should explain how the combination associates. In fact, I think we should almost surely offer more than one version: associating to the left, associating to the right, and (for certain tree-like monoids) associating in a balanced fashion are all potentially useful.

treeowl:
Sorry for the delayed review

chessai:
That is okay, this is valuable feedback. I will make these changes shortly.

Data/Primitive/PrimArray.hs
treeowl: Tree-like or in some cases numerical. Balanced association can reduce floating point errors somewhat, for example.

Data/Primitive/PrimArray.hs
```diff
@@ -467,6 +469,16 @@ sizeofPrimArray :: forall a. Prim a => PrimArray a -> Int
...
+{-# INLINE foldMapPrimArray #-}
+foldMapPrimArray f = foldrPrimArray (\a acc -> acc `mappend` f a) mempty
```
treeowl: This is backwards. Why do you flip the `mappend`?

chessai:
most recent commit does not add balanced foldMap. I think I'd prefer for that to be in a separate PR, as it's a little more complicated.

Data/Primitive/PrimArray.hs
```diff
@@ -467,6 +471,32 @@ sizeofPrimArray :: forall a. Prim a => PrimArray a -> Int
...
+
+-- | Map each element of the primitive array to a monoid, and combine the results.
+--   The combination is left-associated, and the accumulation is lazy.
+foldlMapPrimArray :: forall a m. (Prim a, Monoid m) => (a -> m) -> PrimArray a -> m
+{-# INLINE foldlMapPrimArray #-}
+foldlMapPrimArray f = foldlPrimArray (\acc a -> acc `mappend` f a) mempty
+
+-- | Map each element of the primitive array to a monoid, and combine the results.
+--   The combination is right-associated, and the accumulation is strict, though
+--   this function is not necessarily strict in its result.
```
treeowl: "This function is not necessarily strict in its result" doesn't mean anything whatsoever to me. What I meant needed explanation is that at each step, we force the accumulator value, but we *don't* force the value of `f a`. So if `mappend` is lazy in its first argument, `f a` will not be evaluated. This is probably the correct behavior, but needs better explanation.

changelog.md
```diff
@@ -29,7 +29,9 @@
 
 ## Changes in version 0.6.4.1
 
- * Add instances for the following newtypes from `base`:
+ * Add `foldMapPrimArray` and `foldMapPrimArray'`
```
treeowl: This needs to be updated now that you've added more functions.

chessai:
requested changes attempted, @treeowl

Data/Primitive/PrimArray.hs
```diff
@@ -467,6 +471,65 @@ sizeofPrimArray :: forall a. Prim a => PrimArray a -> Int
...
+--
+-- @mySum = 'Data.Monoid.getSum' '$' 'Data.Monoid.Sum' 10@
+--
+-- @mySum = 10@
+foldMapRPrimArray :: forall a m. (Prim a, Monoid m) => (a -> m) -> PrimArray a -> m
+{-# INLINE foldMapRPrimArray #-}
+foldMapRPrimArray f = foldrPrimArray (\a acc -> f a `mappend` acc) mempty
+
+-- | Map each element of the primitive array to a monoid, and combine the results.
+--   The combination is left-associated, and the accumulation is lazy.
```
treeowl: Example! Connect the left and right versions in docs (even barely).

Data/Primitive/PrimArray.hs
```diff
@@ -467,6 +471,65 @@ sizeofPrimArray :: forall a. Prim a => PrimArray a -> Int
 {-# INLINE sizeofPrimArray #-}
 sizeofPrimArray (PrimArray arr#) = I# (quotInt# (sizeofByteArray# arr#) (sizeOf# (undefined :: a)))
 
+-- | Map each element of the primitive array to a monoid, and combine the results.
+--   The combination is right-associated, and the accumulation is lazy.
+--   An example:
+--
```
treeowl: Use `==== __Example__` to collapse the example. It's an experimental feature, but we use it anyway.

Data/Primitive/PrimArray.hs
chessai: Clarification: Are you asking to provide an example for foldMapLPrimArray, to show left-associativity (for this particular lazy fold)

treeowl:
Both examples are supposed to show associativity. I think they can be much
briefer, TBH. Just give a three-element example application and show how it
unfolds as a single expression.

chessai:
> Both examples are supposed to show associativity. I think they can be much
briefer, TBH. Just give a three-element example application and show how it
unfolds as a single expression.

By 'briefer', do you mean less 'tutorial-like', to assume that the example is illustrative enough sans written explanation?

treeowl:
> By 'briefer', do you mean less 'tutorial-like', to assume that the example is illustrative enough sans written explanation?

Right. This is `primitive`, not `base` or `containers`. Users are expected to know what they're doing. But language is slippery, so "associates to the left" might well evoke opposite meanings in the minds of two different people.

chessai:
> Right. This is primitive, not base or containers. Users are expected to know what they're doing. But language is slippery, so "associates to the left" might well evoke opposite meanings in the minds of two different people.

Makes sense - understood

treeowl:
Aha! I've changed my mind :). I believe we should force each function value
in the strict versions. Laziness here is kind of silly, but more to the
point, I believe the laziness reveals semantic differences in fold
associativity that are only supposed to be performance differences.

chessai:
@treeowl Examples are now simplified and function values are being forced

treeowl:
Humor me and check whether what I said about forcing is really true. I was
going off intuition.

andrewthad:
I like the stuff in the PR, although I dislike the naming convention that @treeowl suggested. I actually prefer `foldrMapPrimArray` to `foldMapRPrimArray`. Anyone else got any thoughts on the color of the bikeshed?

Boarders:
I would still like this and also a foldl for `ByteArray` at some point. Happy to help to make that happen in whatever way I can.

andrewthad:
I'll change the names and merge this. @Boarders Open a separate PR for a foldl on `ByteArray`. You should be able to nearly copy and paste the foldl on PrimArray.

PR #187, opened 2018-07-17
Fix build error by recent ghc-head

TerrorJack:
A recent [commit](https://github.com/ghc/ghc/commit/af9b744bbf1c39078e561b19edd3c5234b361b27) in ghc removed `GHC.Prim.atomicModifyMutVar#` which breaks `Data.Primitive.MutVar`. An implementation of `atomicModifyMutVar#` is provided in `GHC.Exts` instead, so I intend to fix the build error by simply replacing `GHC.Prim` with `GHC.Exts`.

### DIFF
--- a/Data/Primitive/MutVar.hs
+++ b/Data/Primitive/MutVar.hs
@@ -25,7 +25,7 @@ module Data.Primitive.MutVar (
 ) where
 
 import Control.Monad.Primitive ( PrimMonad(..), primitive_ )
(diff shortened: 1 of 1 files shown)


RyanGlScott:
Ah yes, good catch. In general, we ought to be preferring `GHC.Exts` over `GHC.Prim` anyway, since the latter is much more volatile. Thanks for the patch!

cartazio:
is this for master or for 8.6?
(but yeah, GHC.Exts is more stable)

PR #194, opened 2018-08-14
Add new atomic operations

treeowl:
GHC HEAD now offers `atomicModifyMutVar2#` and
`atomicModifyMutVar_#`. Add support for those.

Part of the accepted proposal calls for a new `atomicSwapMutVar#`
primop. That has not yet been implemented, but we can offer a
faithful and pretty efficient implementation.

### DIFF
--- a/Data/Primitive/MutVar.hs
+++ b/Data/Primitive/MutVar.hs
@@ -1,4 +1,10 @@
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE MagicHash, UnboxedTuples, DeriveDataTypeable #-}
+{-# LANGUAGE TypeFamilies, DataKinds, TypeOperators, ConstraintKinds #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE MultiParamTypeClasses, FlexibleInstances #-}
 
 -- |
 -- Module      : Data.Primitive.MutVar
@@ -17,18 +23,34 @@ module Data.Primitive.MutVar (
   newMutVar,
   readMutVar,
   writeMutVar,
-
-  atomicModifyMutVar,
-  atomicModifyMutVar',
   modifyMutVar,
-  modifyMutVar'
+  modifyMutVar',
+
+  atomicModifyMutVar',
+  atomicModifyMutVar,
+  atomicModifyMutVar2,
+  atomicModifyMutVar2Lazy,
+  FirstFieldIs,
+  atomicModifyMutVar'_,
+  atomicModifyMutVar_Lazy,
+  atomicSwapMutVar,
+  casMutVar
 ) where
 
 import Control.Monad.Primitive ( PrimMonad(..), primitive_ )
 import GHC.Exts ( MutVar#, sameMutVar#, newMutVar#,
-                  readMutVar#, writeMutVar#, atomicModifyMutVar# )
+                  readMutVar#, writeMutVar#, atomicModifyMutVar#,
+                  casMutVar# )
 import Data.Primitive.Internal.Compat ( isTrue# )
 import Data.Typeable ( Typeable )
+import GHC.Generics
+-- TODO: Once GHC 8.8 has a base library version, we should
+-- use MIN_VERSION_base here.
+#if __GLASGOW_HASKELL__ >= 807
+import GHC.Exts ( atomicModifyMutVar2#, atomicModifyMutVar_# )
+import GHC.TypeLits (TypeError, ErrorMessage (..))
+import Data.Kind (Constraint)
+#endif
 
 -- | A 'MutVar' behaves like a single-element mutable array associated
 -- with a primitive state token.
@@ -55,8 +77,29 @@ writeMutVar :: PrimMonad m => MutVar (PrimState m) a -> a -> m ()
 {-# INLINE writeMutVar #-}
 writeMutVar (MutVar mv#) newValue = primitive_ (writeMutVar# mv# newValue)
 
--- | Atomically mutate the contents of a 'MutVar'
+-- | Mutate the contents of a 'MutVar'
+modifyMutVar :: PrimMonad m => MutVar (PrimState m) a -> (a -> a) -> m ()
+{-# INLINE modifyMutVar #-}
+modifyMutVar (MutVar mv#) g = primitive_ $ \s# ->
+  case readMutVar# mv# s# of
+    (# s'#, a #) -> writeMutVar# mv# (g a) s'#
+
+-- | Strict version of 'modifyMutVar'
+modifyMutVar' :: PrimMonad m => MutVar (PrimState m) a -> (a -> a) -> m ()
+{-# INLINE modifyMutVar' #-}
+modifyMutVar' (MutVar mv#) g = primitive_ $ \s# ->
+  case readMutVar# mv# s# of
+    (# s'#, a #) -> let a' = g a in a' `seq` writeMutVar# mv# a' s'#
+
+-- | Atomically mutate the contents of a 'MutVar'. This
+-- function is gratuitously lazy. Most applications should use
+-- `atomicModifyMutVar'`, `atomicModifyMutVar2`, or
+-- `atomicModifyMutVar'_`.
 atomicModifyMutVar :: PrimMonad m => MutVar (PrimState m) a -> (a -> (a,b)) -> m b
+-- TODO: Generalize the type the way we do for atomicModifyMutVar2.
+-- The generalization is more annoying to implement and considerably
+-- less powerful, but it can be useful to use two-field records that are not
+-- actual pairs.
 {-# INLINE atomicModifyMutVar #-}
 atomicModifyMutVar (MutVar mv#) f = primitive $ atomicModifyMutVar# mv# f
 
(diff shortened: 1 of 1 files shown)


treeowl:
Note: the weird "`Lazy`" names have a reason. In particular, I believe *very few people* will ever want `atomicModifyMutVar2Lazy` or `atomicModifyMutVar_Lazy`. I want to make it clear that these are *weird* (it's too late to rename `atomicModifyMutVar` itself, but that's pretty weird too). Furthermore, `atomicModifyMutVar2'` would be a very confusing name, since there are *three* different potential ways to be "strict" here. I've made `atomicModifyMutVar2` strict in the function result (the mildest way to be strict). Users can impose additional strictness by using record types with strict fields.

I retain the prime marker on `atomicModifyMutVar'` because it forces a truly polymorphic value. While the value forced by `atomicModifyMutVar2` is polymorphic, it is constrained to be a datatype, which is a reasonable thing to force.

I'm open to general suggestions for name changes, but I don't want to get bogged down too much in those.

treeowl:
@cartazio, @andrewthad, and anyone else who cares: I'm planning to merge this around the same day GHC 8.8-alpha1 shows up. If you want things changed, please say so before then!

treeowl:
Hrmph. I'm going to replace the compatibility shim with one that uses `GHC.Generics.from`. That'll get rid of a lot of headaches, though it will still accept some things the proper version rightly rejects.

treeowl:
@RyanGlScott, I think you might find this interesting and/or have some suggestions.

RyanGlScott:
Sorry for taking a while to look at this.

While most of this looks reasonable, I'm not terribly enthusiastic about `FirstFieldIs`, for several reasons:

1. It involves some fairly intricate type-level machinery that feels rather out-of-place in `primitive`. I've never quite been comfortable with putting `TypeError` in my code that other people intend to use, since they're somewhat brittle and prone to bugginess, as the numerous open tickets in https://ghc.haskell.org/trac/ghc/query?status=!closed&keywords=~CustomTypeErrors witness.
2. Moreover, I don't think it wins us much. As you admit, it's not a perfect safeguard anyway, since it relies on the assumption that you're working on a derived `Generic` instance, which is an assumption that is easily circumvented.
3. There's enough difference in the code between the GHC 8.8 version and the non-8.8 version that I can't help but worry that the conditional API offered here is going to bite someone eventually.

Given that we're in `primitive` anyway, my inclination is to just state what invariants the programmer should uphold and leave it at that.

treeowl:
Perfect safety was never the goal. The goal is to make it easier to avoid breaking the rules accidentally. If you don't like `TypeError`, we could go with a stripped-down `FirstField r ~ a`. I don't think the conditional bits are too likely to cause problems. As long as users are compiling with multiple GHC versions and not using custom generics, the compiler should catch any errors. The alternative compatibility shim is chock-full of `unsafeCoerce fst`, which I don't much care for.

RyanGlScott:
To be honest, I would propose not having a compatibility shim at all, and only expose it on 8.8+. That way, we shouldn't need to deal with `GHC.Generics` at all. (The 8.8 version doesn't make use of it, right?)

To be clear, I do think `FirstFieldIs` has some utility, but I don't think `primitive` is the right library to put it in, following `primitive`'s general philosophy of offering thin abstractions over primops (and I'm hesitant to call `FirstFieldIs` "thin").

treeowl:
What's non-thin about it? Yes, I suppose we could just expose the overly-polymorphic type, but it makes me a bit sad. Should we similarly expose a bogus type for `atomicModifyMutVar`?

RyanGlScott:
> What's non-thin about it?

My biggest concern is that it imposes a `Generic` constraint on the result type of the function, which is a pretty unconventional requirement for `primitive` (or most libraries, for that matter). Typically, `GHC.Generics` operates on an "opt-in" basis, where users can reach for it when they want to opt in to datatype-generic goodies. Here, though, users have no choice. Among other things, this means that there are several data types that simply won't be usable as the result type of the function used to modify the `MutVar`'s contents.

If `Generic` really were bulletproof enough to ensure that no one could ever use an invalid data type as the result type of the function, then I'd be more inclined to pay the price.

A secondary concern is the use of `TypeError`, yes. Maybe I'm being too paranoid here, though, and these combinators are unlikely to be used in sufficiently polymorphic situations to trigger the sorts of bugs that `TypeError` is prone to.

> Yes, I suppose we could just expose the overly-polymorphic type, but it makes me a bit sad.

It is bogus, true, but I don't think one needs to feel bad about it, since `primitive` exposes many, many other things that can be used bogusly.

> Should we similarly expose a bogus type for `atomicModifyMutVar`?

I'm not sure what you mean here. Is the current type for `atomicModifyMutVar` the one's that's bogus?

treeowl:
The current type of `atomicModifyMutVar` is insufficiently polymorphic. The type of the associated primop/former primop is bogus.

RyanGlScott:
> The current type of `atomicModifyMutVar` is insufficiently polymorphic.

Ah, right. But does it have the unsafety that `atomicModifyMutVar2` currently has? If not, would one way to avoid `atomicModifyMutVar2`'s unsafety be to give it the type:

```haskell
atomicModifyMutVar2 :: PrimMonad m => MutVar (PrimState m) a -> (a -> (a,b)) -> m b
```

While it's not as polymorphic as it could be, I would hope that `(a,b)` will work fine for the vast majority of use cases.

treeowl:
I absolutely refuse to give it such a restricted type! I don't want to have to pay for extra pair allocation and maybe more to cram the result into a pair.

RyanGlScott:
OK. In that case, is there a suitable data type that you can concoct to use in its place?

treeowl:
I don't understand the question.

RyanGlScott:
I'm wondering if there is a data type that we can define in `primitive` that we can use as the "canonical" result type for the function passes to `atomicModifyMutVar2`, similarly to how the pair type is the "canonical" result type for `atomicModifyMutVar` currently.

RyanGlScott:
Alright. In that case, ignore my ramblings above—I'd be fine with having the function argument's type simply being `(a -> r)` in the signature for `atomicModifyMutVar2`.

I don't know if I have a strong opinion on what the type signature of `atomicModifyMutVar` should be. I would just say "go ahead and change it", but unfortunately, that would be a breaking change. (If other folks are on board with that idea, I could be easily convinced, though.)

treeowl:
I don't really care too much about `atomicModifyMutVar`. It's not a primop anymore.

treeowl:
Maybe I can have a `saferAtomicModifyMutVar2` that uses the `Generic` stuff? That also has better inference than the completely general version.

RyanGlScott:
Indeed, I think `saferAtomicModifyMutVar2` is worth putting _somewhere_—I'm just left wondering if `primitive` is the right library to call it home.

One point in this idea's favor is that if the we expose both functions, then folks can always fall back on the less safe version if we end up needing to change `FirstFieldIs` and that ends up being incompatible with their use case (for whatever reason that may be).

On the other hand, it does add yet another variant of `atomicModifyWarbleGarble`, which means you'll have to do quite a bit of digging to find out which one you'll actually want to use in your code (perhaps this is a good thing, since there are enough gotchas that one _ought_ to think carefully before choosing).

In other words, I'm on the fence about this. Do others think this plan is reasonable going forward?

cartazio:
i'm looking at this now and i have to agree... 

1) type level programming style safety checks aren't appropriate for primitive, they should live somewhere else, though I DO like this **VERY** clever trick. But I think for primitive lets stick with the Tuple (,) based interface.

cartazio:
I agree with ryan, it does deserve its own package for that generalization,

treeowl:
No, we will not stick with the tuple interface. It's really not general
enough to get optimal performance in many cases. If you don't want cool
type-level stuff, let's stick to a completely unsafe interface.

cartazio:
why can't this generalized version live in its own  little `safe-atomics` package?

treeowl:
Because people expect primitive to give them full access to the power of
the primitives, right? And we quite explicitly don't care about safety
(except for instances of standard classes), leaving that up to the user.
Why should this be an exception to the rule?

cartazio:
mission statement of Primitive is : to support writing simple mutable code that works in both IO and ST style environments, with an primary orientation towards supporting especially the vector package (and friends).

I've not done a good job of communicating/hewwing to that line as clearly as I should in my communications this summer, but thats my stance.

nothing prevents you from packaging up your own techniques and ideas, as edwardk does with his structs package, http://hackage.haskell.org/package/structs. But while we do rebundle some atomics around MutVar and Mvar, primitive really isn't the right place to put this sort of IO/Concurrency optimization trick, especially since theres no performance benefit to a compare and swap loop on single threaded computation.

to restate myself: its a brilliant trick, but it doesn't fit primitive

RyanGlScott:
To reiterate my position on this: I think the unsafe shim definitely warrants inclusion in `primitive` (with the caveat that we should make it clear that it's unsafe).

As for a potential safer version, there definitely is a tension here between offering a safe version in its full generality, which relies on some fairly avant-garde type-level programming features (see https://github.com/haskell/primitive/pull/194#issuecomment-414359212), and offering a safe version monomorphized to a data type (like `(,)`), which would sacrifice performance in certain situations. I don't see a way to meet both of these goals, and at the risk of flaring tensions further, I think we should defer the question of how to best offer a safe version for later.

I realize that the description of this package ("This package provides various primitive memory-related operations.") doesn't really make clear whether David's proposed type for `atomicModifyMutVar2` falls within the scope of `primitive` or not. That being said, it is certainly a departure from most other wrappers in `primitive` in terms of the additional language features it brings into the picture, and as I try to explain in https://github.com/haskell/primitive/pull/194#issuecomment-414359212, I don't feel that these features are battle-tested enough to justify their costs.

That's not to say that I believe `primitive` should shy away from using any GHC-specific tricks altogether. Indeed, searching through `primitive` will yield quite a number of uses of language extensions. It's my belief (at least, my sincere hope) that all of these extensions currently in use have been used enough to understand exactly what benefits they provide and that they don't encumber the API that we expose to users. I was unclear whether `atomicModifyMutVar2` met these requirements, which is why I deferred to others' opinions in https://github.com/haskell/primitive/pull/194#issuecomment-414383133. Given that there has been pushback against this idea, I'm coming to the conclusion that it's out-of-scope for `primitive`.

In light of this, it might be wise to state in greater detail what the scope of `primitive` _is_ exactly. I feel like each person has their own idea of what purpose `primitive` is intended to serve, which might explain why it's easy to get into heated debates about whether something belongs in `primitive` or not.

cartazio:
to restate what i think
mission statement of Primitive is : to support writing simple mutable code
that works in both IO and ST style environments, with an primary
orientation towards supporting especially the vector package (and friends).

I agree with ryan that primitive should adhere to using
stable/well-understood/mature  - interfaces / APIs / GHC Extensions. And I
do think he and I have failed to clearly communicate this to contributors.
And enforce that vision.

andrewthad:
I would like to see this in primitive:

    unsafeAtomicModifyMutVar2 :: MutVar (PrimState m) a -> (a -> r) -> m (a, r)

My preference is that the fully general safe variant of this function live in another package.

cartazio:
@Andrew Martin <andrew.thaddeus@gmail.com> while it should live somewhere,
i dont think it should in primitive,
1) i'm not sure what natural examples where the shaved indirection has a
measurable performance  impact

2) its way way more unsafe that most unsafe things.

cartazio:
(especially since the target memory location will already be in L1 cache
for many of the examples where the indirection lookup overhead is otherwise
a measurable part of the computation)

compare and swap instructions / loops thereof are often the same order of
magnitude or greater cycle count than an L1 lookup..

treeowl:
@cartazio, I don't know why you're so tightly focused on indirection. There's also the cost of allocating the pair nobody needs.

cartazio:
ok, i've looked up the rough cycle counts:
on modern hardware a successful L1 data cache lookup is at most 4 cycles, but a single atomic compare and swap will be 8-50 cycles (theres some variation ).   

the atomic modify family of mutvar/ioref operations are built on top of cmm primops that use compare and swap in a retry loop.

in non-concurrent applications, atomics are always slower than what you should do, and in concurrent applications, the reason for using compare and swap operations in the style of atomic modify is for lock free engineering. (lock free  engineering tends to have worse constant factors, but is easier to compose than lock based algorithms, which tend to have better constant factors).

@treeowl ok, so is it the increase memory pressure, OR the implied heap checks? the tuple certainly wont live past the nursery, so it wont increase the frequency of global gc pauses right? At least in most haskell concurrent applications, the limiting factor for application performance tends to be fraction of time spent in global heap gc.

@treeowl what about we make a version of atomic modify primops that take a result continuation/function with an unboxed tuple arguments/results? I'm open to exploring that. It would address the concern you have with the current api, but still be much much simpler than what we're currently circling.

treeowl:
> @treeowl ok, so is it the increase memory pressure, OR the implied heap checks? the tuple certainly wont live past the nursery, so it wont increase the frequency of global gc pauses right?

No, I don't *think* that's right. As I understand it, we *always* stop the world for GC. There's been some work on trying to avoid that, but my understanding is that it's never been merged (too complicated and too many trade-offs). I'm guessing the remembered sets get hard to handle. So as I understand it, highly concurrent programs currently need to minimize even ephemeral allocation.

> @treeowl what about we make a version of atomic modify primops that take a result continuation/function with an unboxed tuple arguments/results? I'm open to exploring that. It would address the concern you have with the current api, but still be much much simpler than what we're currently circling.

I don't see how that does anything useful whatsoever.

cartazio:
http://hackage.haskell.org/package/atomic-primops by ryan-newton would be a great playground for this approach to be added to it ...

treeowl:
That package seems to be on life support.

cartazio:
you could probably ask ryan newton to take it over and modernize it. (or take the good ideas and evolve a new package!). heck ed also has some crazy stuff thats not right for primitive but are of the same/similar vein in https://github.com/ekmett/concurrent/tree/master/src/Concurrent and related stuff. 



to be clear: I **genuinely** think this  code /  PR is brilliant, like, i had to take a nap after i understood it because it warped my brain with its nugget of clever. I just dont think it belongs in primitive from my perspective / feelings on what the core goals/role/purpose of the primitive package are. 

lots of clever stuff is out there that lives in the Primitive Monad but need not be  included in the primitive package. Vector, ed's various mutable data structure experiments such as `structs`, `concurrent` (as mentioned before), his `transient` code https://github.com/ekmett/transients etc.

cartazio:
i may circle back on adding the other atomics later, but for now i've fixed master to work with both pre and post *mutvar# operations being moved into ghc.exts

PR #196, opened 2018-08-15
Set the stage for UnliftedArray#

treeowl:
The `PrimUnlifted` class, used to implement `UnliftedArray`
operations, is currently based on an `ArrayArray#` representation
of unlifted arrays. But that's set to change, hopefully soon.
I don't think it's too soon to set up the machinery necessary to
allow users to write code *today* that will work once `ArrayArray#`
is gone.

The key here is to add an associated type family to `PrimUnlifted`
that ties each `LiftedRep` type to its `UnliftedRep` counterpart,
and methods that convert back and forth.

### DIFF
--- a/Data/Primitive/Types/PrimUnlifted.hs
+++ b/Data/Primitive/Types/PrimUnlifted.hs
@@ -1,0 +1,302 @@
+{-# language CPP #-}
+{-# language MagicHash #-}
+{-# language TypeFamilies #-}
+{-# language DataKinds #-}
+{-# language DefaultSignatures #-}
+{-# language ScopedTypeVariables #-}
+
+-- |
+-- Module      : Data.Primitive.UnliftedArray
+-- Copyright   : (c) Dan Doel 2016, (c) David Feuer 2018
+-- License     : BSD-style
+--
+-- Maintainer  : Libraries <libraries@haskell.org>
+-- Portability : non-portable
+--
+-- Note: In a previous release, there was a `PrimUnlifted` instance for
+-- `StablePtr`. However, since `StablePtr#` is not of kind `TYPE UnliftedRep`,
+-- and is treated very differently by the garbage collector, this instance was
+-- unsound, and it has been removed.
+--
+-- @since TODO
+
+module Data.Primitive.Types.PrimUnlifted
+ ( PrimUnlifted (..)
+#if __GLASGOW_HASKELL__ < 800
+ , Unlifted
+ , toUnlifted#
+ , fromUnlifted#
+#endif
+ ) where
+
+import GHC.Exts
+
+import Data.Primitive.PrimArray (PrimArray (..), MutablePrimArray (..))
+import Data.Primitive.Array (Array (..), MutableArray (..))
+import Data.Primitive.SmallArray (SmallArray (..), SmallMutableArray (..))
+import Data.Primitive.ByteArray (ByteArray (..), MutableByteArray (..))
+import Data.Primitive.MutVar (MutVar (..))
+import GHC.MVar (MVar(..))
+import GHC.Conc (TVar(..))
+import GHC.Weak (Weak(..))
+import GHC.Conc.Sync (ThreadId(..))
+
+-- | Classifies the types that are just liftings of unlifted pointer
+-- types.
+--
+-- For GHC versions 8.0 and later, instances should define 'Unlift',
+-- 'toUnlifted#', and 'fromUnlifted#'. When compiled with earlier versions,
+-- @Unlifted a@ is always a synonym for 'ArrayArray#', and instances should
+-- define 'toArrayArray#' and 'fromArrayArray#', coercing the internal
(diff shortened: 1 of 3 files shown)


andrewthad:
I agree with this change, and I like this approach to backwards compatibility. Hopefully, I can finish `UnliftedNewtypes` soon and then work on replacing `ArrayArray#`.

treeowl:
@andrewthad, what we really need here, I think, is `UnliftedArray#`. Is there some reason you're waiting with that (simple) feature until you finish the (complicated) unlifted newtype business? Nothing really prevents primitive `ArrayArray#` and `UnliftedArray#` from living side by side for a version or two.

Data/Primitive/UnliftedArray.hs
```diff
@@ -127,90 +130,233 @@ import GHC.Base (runRW#)
...
+  type Unlift a :: TYPE 'UnliftedRep
+#  else
+  type Unlift a :: TYPE 'PtrRepUnlifted
+#  endif
+  type Unlift a = ArrayArray#
+#endif
+
+  -- | Convert a value to its unlifted representation.
+  --
+  -- @since TODO
```
treeowl: We should probably use `DefaultSignatures` here. That means

1. If someone defines `Unlift`, they'll also have to define `toUnlifted#` and `fromUnlifted#`, which seems reasonable.
2. We won't need the `unsafeCoerce#`.

andrewthad:
I guess they could live side-by-side. I hadn't really considered the possibility of doing that, but that should work fine.

Data/Primitive/UnliftedArray.hs
andrewthad: What signature do you have in mind? I don't see what constraint you could possibly provide.

treeowl: I just updated. I *think* I like this way better.

treeowl:
Hrmph... It looks like there's an unfortunate interaction between unlifted types and `DefaultSignatures` in older GHC versions. I guess some extra CPP can fix it.

treeowl:
Ah, I just realized something. Any custom implementation of the new methods for GHC before 8.0 is pretty much guaranteed to be wrong (some silly application of `unsafeCoerce`). So we should only have those methods for 8.0 and above, and make them plain functions in earlier versions. That also simplifies the CPP.

Data/Primitive/Types/PrimUnlifted.hs
```diff
@@ -0,0 +1,274 @@
...
+-- @Unlift a@ is always a synonym for 'ArrayArray#', and instances should
+-- define 'toArrayArray#' and 'fromArrayArray#', coercing the internal
+-- unlifted types to and from 'ArrayArray#'.
+--
+-- Note: 'toArrayArray#' and 'fromArrayArray#' will eventually be
+-- deprecated and removed. They should be used /only/ to define instances
+-- for GHC versions before 8.0.
+class PrimUnlifted a where
+#if __GLASGOW_HASKELL__ >= 800
+  {-# MINIMAL (toArrayArray# | toUnlifted#), (fromArrayArray# | fromUnlifted# ) #-}
```
treeowl: Should we change this to `{-# MINIMAL toUnlifted#, fromUnlifted# #-}`? That will produce a technically spurious warning, but we really do want users to get used to defining these methods for 8.0 and up.

andrewthad: I agree that `{-# MINIMAL toUnlifted#, fromUnlifted# #-}` is better.

treeowl:
@andrewthad, I've shaken things up enough now that I'd appreciate a second full review. One specific question: should I rename `Unlift` to `Lower` or something? One thing to consider, perhaps, is that [`monad-unlift`](http://hackage.haskell.org/package/monad-unlift-0.2.0/docs/Control-Monad-Trans-Unlift.html) exports an `Unlift` type. I don't know if there's much potential for name clashes there.

treeowl:
Ah, right..... I think we currently (pretty much?) have to export the class and legacy methods from `Data.Primitive.UnliftedArray` to avoid breaking all existing user instances. Gross! What I think we *ultimately* want is to re-export just the class from `Data.Primitive.UnliftedArray`, and perhaps to export the class and non-legacy methods from `Data.Primitive.Types`. Hrmmm.....

treeowl:
@andrewthad, I think this is ready to merge, except for the question of the name of the `Unlift` type family. Can you give an opinion about that?

andrewthad:
To me, `Unlifted` is the most natural choice. The other names involved in this class (`PrimUnlifted`, `toUnlifted`, `fromUnlifted`) all use the word `Unlifted`, and I think the type family should too.

treeowl:
I definitely agree it's the most natural. I suppose if there's a name clash in the future we can work it out then. I'll make the change. Once you've added a note about the `UnliftedRep (StablePtr a)` replacement (which I'm really hoping you'll do in the next few minutes), I'll rebase this on top of that and merge.

Issue #195, opened 2018-08-15
PrimUnlifted StablePtr looks sketchy


treeowl:
We currently have `PrimUnlifted (StablePtr a)`, but that's a bit troubling. `StablePtr# a` has kind `TYPE 'AddrRep` rather than `TYPE 'UnliftedRep`, which means that even once we have `UnliftedArray#` we won't be able to put a `StablePtr#` into one without `unsafeCoerce#`. Is that coercion safe? I don't personally have a clue; maybe someone who knows more about stable pointers can answer that question.

treeowl:
@simonmar, I suspect you're the one best positioned to answer this question.

treeowl:
That was my concern, but I don't know enough about stable pointers. What does the GC do when it collects the contents of an `ArrayArray#`? What will happen if it encounters a `StablePtr#` in one of the slots? In particular, will it "notice" that the `StablePtr#` is a `StablePtr#` and just do the right thing (in which case we can keep using the instance) or will it crash?

treeowl:
@bgamari, you may know something about this too.

treeowl:
All righty, then. We'd best get this fixed. Thanks, all.

andrewthad:
This instance causes a [compile-time crash for ghcjs 8.4](https://github.com/ghcjs/ghcjs/issues/665). I going to put up a PR removing it.

PR #202, opened 2018-08-17
remove PrimUnlifted instance for StablePtr

andrewthad:
Resolves https://github.com/haskell/primitive/issues/195

### DIFF
--- a/Data/Primitive/UnliftedArray.hs
+++ b/Data/Primitive/UnliftedArray.hs
@@ -42,6 +42,10 @@
 -- 'MutableUnliftedArray' are parameterized by the type of arrays they contain, and
 -- the coercions necessary are abstracted into a class, 'PrimUnlifted', of things
 -- that are eligible to be stored.
+--
+-- Note: In a previous release, there was a `PrimUnlifted` instance for
+-- `StablePtr`. However, since `StablePtr#` does is not of kind `TYPE UnliftedRep`,
+-- this instance was unsound, and it has been removed.
 
 module Data.Primitive.UnliftedArray
   ( -- * Types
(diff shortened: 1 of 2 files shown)


treeowl:
I think it would be a good idea to add a note explaining why we don't have the instance, to (hopefully) prevent someone from trying to put it back.

hvr:
stating the obvious: don't forget to major bump to 0.7.0

treeowl:
@hvr, I think I disagree. Yes, removing an instance should trigger a major version bump, in theory. However, in this case, any version of any package that actually relies on this instance should be considered badly broken, as it will lead to segfaults. Such packages should get minor (bugfix) releases and their broken versions should be deprecated. Isn't it better to fail to compile than to compile to segfaulting code?

andrewthad:
I've added a note to the changelog and to the module about the removed instance. I'm punting on the question of a major version for now, but my preference would be for this not to be a major version bump.

andrewthad:
Spurious travis failure on GHC 8.0.2.

hvr:
@treeowl you'd be conflating concerns if you do that; you'd be abusing the versioning policy to intentionally trigger compile failures; which in turn will force Hackage Trustees to go around and fix the upper bounds to `< 0.6.5` for the packages affected to rectify the metadata and make the packages compile again; iow, we're going to have to spend additional effort to work against your intentional compile failure sabotage act... ;-) and the end-result will be basically the same as if you'd hadn't deliberately violated the PVP to begin with; except that you've costed us more effort.

hvr:
@treeowl instead, I'd suggest a different course of action which is more principled: 

- blacklist `primitive-0.6.4` so that the solver cannot access it anymore
- release `primitive-0.6.4.1` whose only change relative to the redacted `primitive-0.6.4` is that it removes the instance

Hackage trustees will still have to go around fixing up affected packages who started depending on that instance; but the big difference is that packages won't be able anymore to use `primitive-0.6.4` as it's not accessible to the solver anymore *at all*.

cartazio:
If we’re removing a type class instance it’s a major versionbump.

treeowl:
@hvr the principled course of action you propose is exactly what I was talking about. There's no major version bump there!

PR #208, opened 2018-08-20
interleave

strake:


treeowl:
`noDuplicate#` didn't have a token-polymorphic type until GHC 8.2, when I generalized it to fix a bug in lazy `ST`. You'll have to `unsafeCoerce` your way to glory in earlier versions.

treeowl:
I added a commit to fix that. Feel free to use it or not.

strake:
LGTM

treeowl:
The build is broken on pre-AMP versions. We need to import `Control.Applicative` for those, or switch from `*>` to `>>`.

strake:
@RyanGlScott Done, PTAL

strake:
@RyanGlScott Done, PTAL

changelog.md
```diff
@@ -1,3 +1,7 @@
+## Changes in version 0.?.?.?
```
RyanGlScott: There's already a "Changes in version 0.?.?.?" section below. Put the mention of `unsafeInterleave` in that section instead.

treeowl: The likely reason for confusion is that we already have two, one for a bug-fix release and one for the next major or minor release.

RyanGlScott: Oh dear, I didn't notice that! In that case, I'm not exactly sure myself where this mention of `unsafeInterleave` should go...

treeowl:
It should go in the top-most mystery section. It's a new feature, so if there's a bug-fix-only release it won't include this.

strake:
Done

RyanGlScott:
LGTM. Thanks!

PR #209, opened 2018-08-24
Use newly-exported StableName constructor

treeowl:
GHC 8.6 will export the `StableName` constructor from
`GHC.StableName`, meaning we won't have to pull an ugly
`unsafeCoerce` trick to get at its guts.


treeowl:
We need to wait with this for a bit. I'm hoping @bgamari will bump the base library version for 8.6 by a minor version so we can use this there.

treeowl:
Ah, it seems I misunderstood alpha versioning. Never mind, @bgamari .

PR #220, opened 2018-12-12
Fix documentation comment for copyMutableArray.

theindigamer:
Related question on [StackOverflow](https://stackoverflow.com/questions/53734159/subtle-difference-between-copymutablearray-and-copymutablearray). I hope such a small fix doesn't have to go through the libraries mailing list :sweat_smile: .

### DIFF
--- a/Data/Primitive/Array.hs
+++ b/Data/Primitive/Array.hs
@@ -235,7 +235,8 @@ copyArray !dst !doff !src !soff !len = go 0
 -- same and the ranges overlap, the copy still works as expected, as if the
 -- region is first copied to a temporary buffer.
 --
--- Note: The order of arguments is different from that of 'copyMutableArray#'.
+-- Note: The order of arguments is different from that of 'copyMutableArray#'. The primop
+-- has the source first while this wrapper has the destination first.
 copyMutableArray :: PrimMonad m
           => MutableArray (PrimState m) a    -- ^ destination array
           -> Int                             -- ^ offset into destination array
--- a/Data/Primitive/Array.hs
+++ b/Data/Primitive/Array.hs
@@ -231,8 +231,11 @@ copyArray !dst !doff !src !soff !len = go 0
          | otherwise = return ()
 #endif
 
--- | Copy a slice of a mutable array to another array. The two arrays may
--- not be the same.
+-- | Copy a slice of a mutable array to another array. If the two arrays are the
+-- same and the ranges overlap, the copy still works as expected, as if the
+-- region is first copied to a temporary buffer.
+--
+-- Note: The order of arguments is different from that of 'copyMutableArray#'.
 copyMutableArray :: PrimMonad m
           => MutableArray (PrimState m) a    -- ^ destination array
           -> Int                             -- ^ offset into destination array


andrewthad:
Thanks, and I agree this is worth clarifying in the docs. This has even [tripped up the maintainers](https://github.com/haskell/primitive/pull/116/files) before.

cartazio:
1) awesome documentation fix!

2) i'mnot sure if i agree that the change to the documented specification/semantics should have been merged in too. I'll take a look at the code and think abut the design implications, but that part really shouldn't have been merged in without some discussion first

treeowl:
Carter, it's just a thin wrapper. There'd better be a good reason not to
document it as such.

cartazio:
ok, that new version of the comment is correct in base 4.12, but i'll have to git blame to track down WHY the comment said they can't overlap before

cartazio:
ok, it was added in 2011, and the comment was the same then 

```

-- | Copy a slice of an immutable array to a mutable array.
copyArray :: PrimMonad m
          => Array a                         -- ^ source array
          -> Int                             -- ^ offset into source array
          -> MutableArray (PrimState m) a    -- ^ destination array
          -> Int                             -- ^ offset into destination array
          -> Int                             -- ^ number of elements to copy
          -> m ()
{-# INLINE copyArray #-}
#if __GLASGOW_HASKELL__ >= 702
copyArray (Array src#) (I# soff#) (MutableArray dst#) (I# doff#) (I# len#)
  = primitive_ (copyArray# src# soff# dst# doff# len#)
#else
copyArray !src !soff !dst !doff !len = go 0
  where
    go i | i < len = do
                       x <- indexArrayM src (soff+i)
                       writeArray dst (doff+i) x
                       go (i+1)
         | otherwise = return ()
#endif

-- | Copy a slice of a mutable array to another array. The two arrays may
-- not be the same.
copyMutableArray :: PrimMonad m
          => MutableArray (PrimState m) a    -- ^ source array
          -> Int                             -- ^ offset into source array
          -> MutableArray (PrimState m) a    -- ^ destination array
          -> Int                             -- ^ offset into destination array
          -> Int                             -- ^ number of elements to copy
          -> m ()
{-# INLINE copyMutableArray #-}
#if __GLASGOW_HASKELL__ >= 702
copyMutableArray (MutableArray src#) (I# soff#)
                 (MutableArray dst#) (I# doff#) (I# len#)
  = primitive_ (copyMutableArray# src# soff# dst# doff# len#)
#else
copyMutableArray !src !soff !dst !doff !len = go 0
  where
    go i | i < len = do
                       x <- readArray src (soff+i)
                       writeArray dst (doff+i) x
                       go (i+1)
         | otherwise = return ()
#endif
```
so for overlapping arrays in pre 7.6 ghc, the new comment is wrong, and we currently build as far back as 7.4!

cartazio:
Either way, thanks for the doc fix @theindigamer! I'll do some more investingation of semantics over time in ghc land and fix that part up

cartazio:
hrm, the overlap being allowed is only mentioned in the ghc-prim docs as of 0.4 for ghc 7.10 , the 0.3.1 docs for 7.8 dont mention the overlap. 

the docs seem to mention the 7.6 and older implementation had some bugs, though i'm having trouble finding it on track, and it looks like the current CPP *allows* using the 7.6 primop ... which might be a typo?

cartazio:
yup, it looks like it was a typo in allowing = 7.6, i'll fix that in master :)

cartazio:
fix will include making the new documenation true! :) 

thank you to @theindigamer  for making it clear that the old docs reflected a weakness in the fall back implementation

cartazio:
https://github.com/haskell/primitive/commit/1de2104d12a1a937641d0b3c87307cae22007965 is my temporary fix on maste

PR #222, opened 2018-12-14
For GHC 7.6 and up, use GHC.Prim.setByteArray#

andrewthad:
For GHC 7.6 and up, use GHC.Prim.setByteArray# instead of the FFI to fill byte arrays with Int8 and Word8. This also adds some tests to confirm that the behavior has not changed.

### DIFF
--- a/Data/Primitive/Types.hs
+++ b/Data/Primitive/Types.hs
@@ -66,6 +66,12 @@ import Numeric
 
 import qualified Foreign.Storable as FS
 
+#if __GLASGOW_HASKELL__ >= 706
+import GHC.IO (IO(..))
+import qualified GHC.Prim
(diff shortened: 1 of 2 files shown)


andrewthad:
Travis failures look spurious.

cartazio:
@andrewthad  this and 1-2 other things would be great to apply to master after i get the 0.7 stuff green so we can include it there

cartazio:
lets aim for this to land in a soon 0.7.1 :), though i'd like drop having the vendored qc-classes for faster builds first

andrewthad:
I've tweaked the CPP to only use the `setByteArray#` primop when building on GHC 7.10 and higher. This is ready to be merged.

PR #224, opened 2018-12-15
Read arrays more flexibly

### LINKED ISSUE #223: Read Array instance fails in record context
When putting an Array into a record field, the show/read roundtrip fails. The following program

```
import Data.Primitive.Array

data Record = Record { field :: Array Int }
  deriving (Read, Show)

r :: Record
r = Record $ fromListN 0 []

main :: IO ()
main = print (read (show r) :: Record)
```

when run produces
```
λ> main
*** Exception: Prelude.read: no parse
```

treeowl:
Use the `ReadPrec` mechanism instead of the `ReadS` one. Allow reading
"`OverloadedList` style" and "`OverloadedString`" style, along with
`fromList` and `fromListN` style.

Fixes #223

### DIFF
--- a/Data/Primitive/Array.hs
+++ b/Data/Primitive/Array.hs
@@ -68,11 +68,16 @@ import GHC.Exts (runRW#)
 import GHC.Base (runRW#)
 #endif
 
+import Text.Read (Read (..), lexP, parens, prec)
+import Text.Read.Lex (Lexeme (Ident))
+import Text.ParserCombinators.ReadPrec (ReadPrec, readPrec_to_S, readS_to_Prec)
+import qualified Text.ParserCombinators.ReadPrec as ReadPrec
 import Text.ParserCombinators.ReadP
 
 #if MIN_VERSION_base(4,9,0) || MIN_VERSION_transformers(0,4,0)
 import Data.Functor.Classes (Eq1(..),Ord1(..),Show1(..),Read1(..))
 #endif
+import Control.Monad (liftM2)
 
 -- | Boxed arrays
 data Array a = Array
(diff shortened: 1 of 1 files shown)


Data/Primitive/Array.hs
```diff
@@ -783,28 +787,63 @@ instance Show1 Array where
...
+-- can see very quickly when we've run into a problem. We should also get
+-- a slight efficiency boost by going through the string just once.
+lexTag :: ReadP Tag
+lexTag = do
+  string "fromList"
+  s <- look
+  case s of
+    'N':c:_
+      | '0' <= c && c <= '9'
+      -> fail "" -- We have fromListN3 or similar
```
treeowl: The old implementation would accept `"fromListN3 [1,2,3]"`, but that was wrong.

treeowl:
I'm pretty sure this is what we want `Read` to look like; it's forgiving and should also be *reasonably* efficient (we could do better in the `fromListN` case if lists were read uniformly, but they're not, and I don't think it's worth rewrite rules to work around that). Do we want to change the `Show` instance here? I don't care too terribly much.

treeowl:
Lots of spurious failures going on. Maybe a temporary server problem? Going to have to poke them a few times and see what happens.

skeuchel:
- [ ] Fix the Read/Read1 instances for SmallArray too since they are affected by the same issue.
- [ ] Fix the testsuite for Show/Read laws which is suffering from two copy&paste errors 
https://github.com/haskell/primitive/blob/f2ae1b7f6770000a4744a356539a8a35a35071f2/test/main.hs#L85-L89
https://github.com/haskell/primitive/blob/f2ae1b7f6770000a4744a356539a8a35a35071f2/test/main.hs#L111-L113 
- [ ] In fact, the Read instance for ByteArray doesn't even exist yet. Add it and also the one for PrimArray (and UnliftedArray).

Data/Primitive/Array.hs
skeuchel: Just a side note: The Haskell language report also allows non-ascii unicode decimal digits to be used in number literals. But ghc/base don't.

treeowl:
@cartazio, the complexity seems quite manageable to me, especially since we'll probably never be *forced* to change any of this, and I really like `Read` instances to be tolerant. Computationally, this should also be very cheap: I was careful to defer all decisions to the last possible character.

Data/Primitive/Array.hs
treeowl: @skeuchel, I suppose I could use `isAlphaNum` there, but as long as `base` doesn't accept non-ASCII numerals, we don't really have to. I don't feel too strongly about it. `isAlphaNum` is a tad more expensive than this test. I'll leave that decision up to others, if we end up continuing to support `"fromListN"` at all here.

cartazio:
David, i guess you're likely right, but i'll nedd take some time to
convince myself (it is certainly better from a hacky debuggying eperience,
which is what we ultimately use show for)

andrewthad:
I like the forgiving `Read` instances. I haven't actually reviewed the code, but I'm in favor of the spirit of it.

RyanGlScott:
I, too, like the approach here.

It would be good to get this resolved soon, since [`master` is failing to build](https://travis-ci.org/haskell/primitive/builds/489247980) due to this issue.

cartazio:
i'll add this to my review and merge queue for this week 
cc @hvr  for additional feedback

Data/Primitive/Array.hs
cartazio: hex and binary literals are supported in recent ghc

cartazio: @hvr fromListN reading support seems like a good idea, decimal only or should we also allow hex/binary literals?

cartazio:
i need to take a day to thinka bout decimal vs nondecimal lengths,  i could imagine a universe where hex would be handy, but i'm not sure if that ever matters

PR #225, opened 2019-01-05
Improve foldrByteArray

Bodigrim:
1. Replace one multiplication per array's element by a single division per call.
2. Add `INLINE` pragma. This is crucial to get `sizeOf (undefined :: a)` specialized, allowing GHC to replace `quot` by a bitmask in CMM.
3. Add a test case.


Bodigrim:
Travis failure is unrelated to the change and happens because recent `quickcheck-classes-0.6.0.0` introduced more `Show/Read` laws, which existing instances in `primitive` fail to satisfy.

PR #231, opened 2019-03-23
Adapt to MonadFail-related changes in base-4.13

RyanGlScott:
`base-4.13` will finally complete the `MonadFail` proposal by removing the `fail` method from `Monad`. This adapts `primitive` to that change by guarding the existing `fail` implementations in `primitive` (for `Array` and `SmallArray`) behind CPP, and defining `MonadFail` instances for `Array` and `SmallArray` going forward.


cartazio:
Ill review

andrewthad:
I've looked over this, and I can confirm that that it works as expected. Merging.

PR #234, opened 2019-04-26
Bump base bounds

vmchale:
This bumps the bounds so that `primitive` can be built with GHC 8.8.1 now that all is fixed.

### DIFF
--- a/bench/primitive-benchmarks.cabal
+++ b/bench/primitive-benchmarks.cabal
@@ -43,7 +43,7 @@ benchmark bench
     PrimArray.Compare
     PrimArray.Traverse
   build-depends:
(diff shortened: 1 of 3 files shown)


cartazio:
@vmchale 0.6.1.2 should build with 8.8 now (i did a relaxation of its constraints, )

i'm preparing a patch set for 0.6.4, the challenge there is accounting for mkaing sure its just a bug fix bump vs a major version change, current master is 0.7 

I will do a monad fail backport for 0.6.4 too though

chessai:
this should also bump the benchmark and test suite, right?

chessai:
thanks

cartazio:
i've a patch for this that will appear on master shortly

PR #235, opened 2019-05-01
Support GHC 8.8

fumieval:
Pardon an automatically-posted pull request; this is a part of [Operation Vanguard](https://github.com/haskell-vanguard/haskell-vanguard). Note that this does not guarantee that tests and benchmarks are buildable.

### DIFF
--- a/bench/primitive-benchmarks.cabal
+++ b/bench/primitive-benchmarks.cabal
@@ -43,7 +43,7 @@ benchmark bench
     PrimArray.Compare
     PrimArray.Traverse
   build-depends:
-      base >= 4.8 && < 4.13
+      base >= 4.8 && < 4.14
     , primitive
     , deepseq
     , ghc-prim
--- a/primitive.cabal
+++ b/primitive.cabal
@@ -55,7 +55,7 @@ Library
         Data.Primitive.Internal.Compat
         Data.Primitive.Internal.Operations
 
-  Build-Depends: base >= 4.5 && < 4.13
+  Build-Depends: base >= 4.5 && < 4.14
                , ghc-prim >= 0.2 && < 0.6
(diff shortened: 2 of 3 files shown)


Data/Primitive/Array.hs
```diff
@@ -702,8 +702,10 @@ instance Monad Array where
          *> fill (off + lsb) sbs smb
 
 #if !(MIN_VERSION_base(4,13,0))
+#if !MIN_VERSION_base(4,13,0)
```
chessai: this is redundant

Data/Primitive/Array.hs
```diff
@@ -702,8 +702,10 @@ instance Monad Array where
...
   fail = Fail.fail
 #endif
+#endif
```
chessai: this is redundant

Data/Primitive/SmallArray.hs
```diff
@@ -810,8 +810,10 @@ instance Monad SmallArray where
        *> fill (off + length sb) sbs smb
 
 #if !(MIN_VERSION_base(4,13,0))
+#if !MIN_VERSION_base(4,13,0)
```
chessai: this is redundant

chessai:
i'm not sure we'd want to take this in favour of #234, which is already queued up for the 0.7 release. though, in #234 we probably also want to bump the version of base in the benchmarks/test suite

fumieval:
Sorry, my script went in a wrong way; fixed redundant changes. I didn't realise there was already a pull request.

cartazio:
theres already support on hackage 0.6.1.3, and a matching tag on git, 

for managing a package set, master is never going to be the stable interface ... 

as for master, i've a patch that just needs to be rebased ontop of the recent docfix

PR #236, opened 2019-05-01
remove all imports of GHC.Prim (GHC.Prim ==> GHC.Exts). resolves #197

### LINKED ISSUE #197: Quit importing GHC.Prim
There are still one or two modules that import `GHC.Prim`. We should *never* do this. The approved place to grab the primitive stuff is `GHC.Exts`. Among other things, primitives can *go away* and be replaced by compatibility functions that will not be in `GHC.Prim`.

chessai:
this builds on my machine with GHCs 7.4.2 through 8.6.4.

### DIFF
--- a/Data/Primitive/Array.hs
+++ b/Data/Primitive/Array.hs
@@ -30,7 +30,10 @@ module Data.Primitive.Array (
 import Control.Monad.Primitive
 
 import GHC.Base  ( Int(..) )
-import GHC.Prim
+import GHC.Exts
+#if (MIN_VERSION_base(4,7,0))
+  hiding (toList)
+#endif
 import qualified GHC.Exts as Exts
 #if (MIN_VERSION_base(4,7,0))
 import GHC.Exts (fromListN, fromList)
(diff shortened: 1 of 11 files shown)


andrewthad:
Agreed. It is important for this to happen. There's no reason `primitive` should disregard the admonition offered in `GHC.Prim`.

Data/Primitive/Array.hs
```diff
@@ -30,7 +30,10 @@ module Data.Primitive.Array (
...
 
 import GHC.Base  ( Int(..) )
-import GHC.Prim
+import GHC.Exts
+#if (MIN_VERSION_base(4,7,0))
+  hiding (toList)
+#endif
 import qualified GHC.Exts as Exts
 #if (MIN_VERSION_base(4,7,0))
 import GHC.Exts (fromListN, fromList)
```
RyanGlScott: There's a couple of spots in the code where we already import `GHC.Exts` in the same module, such as in this line. It would be good to consolidate these imports with the newly-created `GHC.Exts` imports that this patch introduces.

treeowl:
We should really go a bit further, I think. `GHC.Base` isn't a great module to grab things from either, with few exceptions. `GHC.Exts` is always better when possible.

PR #237, opened 2019-05-09
Remove Cabal flag hack in primitive-benchmarks

RyanGlScott:
It turns out that we can get away without needing an ugly Cabal flag in `primitive-benchmarks` by simply using `haskell-ci` more intelligently. I only added this flag in the first place since I was under the false impression that `haskell-ci` would _always_ pass the `--enable-benchmarks` flag if there was a benchmark suite, but I have since learned that that is not true.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@
 #
 # For more information, see https://github.com/haskell-CI/haskell-ci
 #
-# version: 0.2
+# version: 0.2.1
 #
 language: c
 dist: xenial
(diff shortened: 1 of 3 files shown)


RyanGlScott:
As observed [here](https://github.com/haskell-CI/haskell-ci/issues/273#issuecomment-490973876), there appears to be a bad interaction in `haskell-ci` between the `ghc-head` and `jobs-selection=any` flags. Perhaps we should wait until that bug is fixed before landing this patch.

RyanGlScott:
Closing for now, due to the reasons mentioned in https://github.com/haskell/primitive/pull/237#issuecomment-490982863.

PR #246, opened 2019-05-31
add Eq instances for MutableByteArray and MutablePrimArray

andrewthad:


RyanGlScott:
This probably deserves a mention in the changelog, but other than that, LGTM.

andrewthad:
Done.

cartazio:
@RyanGlScott @andrewthad  .... i'd like to point out that Eq is usually not Reference/Ptr equality, its "same contents" equality ... are we sure we want to do that?

RyanGlScott:
For a mutable data structure like `MutableByteArray`, I think pointer equality makes sense for its `Eq` instance, especially since that is a convention that other mutable data types like `IORef`, `STRef`, and `MVar` adhere to in their `Eq` instances.

andrewthad:
As Ryan points out, this is already done pervasively throughout base.  Also, it’s the only possible instance.

cartazio:
ok, you're both correct, but spelling it out helps a lot! (for me when i'm
slow, and for being able to circle back later :) )

andrewthad:
I’ll add some documentation to the instance haddocks as well.  It can’t hurt to clarify there.

PR #247, opened 2019-05-31
add clonePrimArray and cloneMutablePrimArray

andrewthad:


andrewthad:
I've implemented your suggestion and reopened at https://github.com/haskell/primitive/pull/271 because of a need to rebase.

Issue #249, opened 2019-06-02
Missing instances for Prim WordPtr and Prim IntPtr


ekmett:
What it says in the title.

cartazio:
good catch, thats a must for 0.7.1

PR #250, opened 2019-06-03
Add Prim instances for WordPtr and IntPtr

andrewthad:
Resolves https://github.com/haskell/primitive/issues/249

### DIFF
--- a/Data/Primitive/Types.hs
+++ b/Data/Primitive/Types.hs
@@ -29,7 +29,8 @@ module Data.Primitive.Types (
 import Control.Monad.Primitive
 import Data.Primitive.MachDeps
 import Data.Primitive.Internal.Operations
-import Foreign.Ptr (IntPtr(IntPtr),WordPtr(WordPtr))
+import Foreign.Ptr (IntPtr,intPtrToPtr,ptrToIntPtr)
+import Foreign.Ptr (WordPtr,wordPtrToPtr,ptrToWordPtr)
 import Foreign.C.Types
 import System.Posix.Types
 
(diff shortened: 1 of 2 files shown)


RyanGlScott:
I don't think `GeneralizedNewtypeDeriving` is going to work for `IntPtr` and `WordPtr` on all versions of GHC that `primitive` supports, since the constructors for those data types have only been exported since GHC 8.2 (which is a prerequisite for GND).

andrewthad:
I've rewritten the instances so that they should work with older GHCs.

Data/Primitive/Types.hs
```diff
@@ -390,6 +392,36 @@ deriving instance Prim CTimer
...
+  indexByteArray# a i = ptrToWordPtr (indexByteArray# a i)
+  readByteArray# a i s0 = case readByteArray# a i s0 of
+    (# s1, p #) -> (# s1, ptrToWordPtr p #)
+  writeByteArray# a i wp = writeByteArray# a i (wordPtrToPtr wp)
+  setByteArray# a i n wp = setByteArray# a i n (wordPtrToPtr wp)
+  indexOffAddr# a i = ptrToWordPtr (indexOffAddr# a i)
+  readOffAddr# a i s0 = case readOffAddr# a i s0 of
+    (# s1, p #) -> (# s1, ptrToWordPtr p #)
+  writeOffAddr# a i wp = writeOffAddr# a i (wordPtrToPtr wp)
+  setOffAddr# a i n wp = setOffAddr# a i n (wordPtrToPtr wp)
```
RyanGlScott: Given the large amount of code repetition here, perhaps we should factor this amount into a CPP macro? The existing `derivePrim` is _almost_ what we want, although there are enough differences between the code you're using here and `derivePrim` where I think it would make sense for this to exist as its own macro.

andrewthad: With some clever internal use of pattern synonyms, we could actually reuse `derivePrim` as is. But I think it's better to just have a different macro as you suggest. I'll look into this.

Issue #262, opened 2020-02-29
add st.lazy instances after we drop  < 8.2 support in the future?


cartazio:
sooo, i hate to be the bearer of bad news, but I think this shouldn't have been merged in at present.

until we no longer support GHC older than 8.2, cf https://gitlab.haskell.org/ghc/ghc/commit/2f5cb3d44d05e581b75a47fec222577dfa7a533e (and ticket https://gitlab.haskell.org/ghc/ghc/issues/13615), any code with the lazy st monad is WILDLY thread/concurrency unsafe

atm merged in via 
PR https://github.com/haskell/primitive/pull/254
regarding issue https://github.com/haskell/primitive/issues/24

other options: 
a)whatever, its fine
b) cpp it to only be exposed on ghc >= 8.2

Avi-D-coder:
I'd argue for CPP. 8.2 is quite old at this point.

andrewthad:
I think CPP is the best option here.

chessai:
Sorry about that! Forgot about those issues with old GHCs. The fix with CPP has been merged.

cartazio:
Unless I read the links I shared on this ticket incorrectly, the cpp addition y’all merged in is wrong. Part of that is my fault for writing the wrong title on this ticket.  :))

cartazio:
its my fault  communication wise, i'll fix it up

Avi-D-coder:
Oops I miss understood.

cartazio:
It’s fine.  But this it’s also why I like to take my time on this stuff ;) 
It takes time to make sure things are correct and that folks are on the same page. 

Wouldn’t mind other folks looking at the commit and doc data  to double check

cartazio:
fixed in master

PR #263, opened 2020-02-29
Remove Lazy ST instance on GHC 8.2

Avi-D-coder:
Per #262


chessai:
Thanks!

cartazio:
Ummm... this patch is wrong.  I think it should be >= 8.2 ...


Or was I misreading the links I shared?
I feel Like this was hair trigger merged

PR #264, opened 2020-03-01
Remove Lazy ST instance on GHC >= 8.2

Avi-D-coder:
closes: #262

### DIFF
--- a/Control/Monad/Primitive.hs
+++ b/Control/Monad/Primitive.hs
@@ -210,7 +210,7 @@ instance PrimBase (ST s) where
   internal (ST p) = p
   {-# INLINE internal #-}
 
-#if __GLASGOW_HASKELL__ > 802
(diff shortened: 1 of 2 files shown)


Avi-D-coder:
@cartazio How is does this look right?

Avi-D-coder:
@cartazio does this look right?

cartazio:
Yeah, I’ll apply it on top of some patches I have locally.

cartazio:
Also I like to do patches ff-only and sometimes I have a stack of patches
locally I want to apply :)

cartazio:
merged in (essentially the same as this patch, but with a URI link about the commit in question)

cartazio:
@Avi-D-coder  i'll add an attribution to you as "Avi Dessauer" for the change log, is that the correct name to use?

Avi-D-coder:
@cartazio  Yep

PR #269, opened 2020-06-03
Ensure that HsFFI.h is #include'd first

bgamari:
Otherwise things may break on Windows. For instance,
you may get things like,
```
   includes\stg\Types.h:26:9: error:
     warning: #warning "Mismatch between __USE_MINGW_ANSI_STDIO
     definitions. If using Rts.h make sure it is the first header
     included." [-Wcpp]
```

### DIFF
--- a/cbits/primitive-memops.h
+++ b/cbits/primitive-memops.h
@@ -1,9 +1,11 @@
 #ifndef haskell_primitive_memops_h
 #define haskell_primitive_memops_h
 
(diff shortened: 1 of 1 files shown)


chessai:
Jeez, how old is this bug? Weird.

Also, can't hurt to merge. Even though I don't think any primitive maintainer is able to test other than you. We should probably set up windows CI for primitive.

Issue #267, opened 2020-05-05
NFData instance for ByteArray


Boarders:
For the purposes of using benchmarking with criterion it would be nice if `ByteArray` had an `NFData` instance (so that one can use the [env function](https://hackage.haskell.org/package/criterion-1.5.6.2/docs/Criterion-Main.html#v:env)). Is there a fundamental reason why this doesn't exist?

treeowl:
No, this seems eminently reasonable.

cartazio:
Agreed.

Boarders:
Added a PR [here](https://github.com/haskell/primitive/pull/270) but it is a bit naive.

PR #270, opened 2020-06-04
Add NFData instance (#267)

Boarders:
This adds an NFData instance. It is quite naive so I welcome feedback on important considerations (either with versions or just using `rwhnf`).

### DIFF
--- a/Data/Primitive/ByteArray.hs
+++ b/Data/Primitive/ByteArray.hs
@@ -54,6 +54,7 @@ module Data.Primitive.ByteArray (
 
 import Control.Monad.Primitive
 import Control.Monad.ST
+import Control.DeepSeq
 import Data.Primitive.Types
 
 import Foreign.C.Types
@@ -71,6 +72,7 @@ import Data.Typeable ( Typeable )
 import Data.Data ( Data(..) )
 import Data.Primitive.Internal.Compat ( isTrue#, mkNoRepType )
 import Numeric
+
 
 #if MIN_VERSION_base(4,9,0)
 import qualified Data.Semigroup as SG
@@ -97,6 +99,12 @@ data ByteArray = ByteArray ByteArray# deriving ( Typeable )
 -- | Mutable byte arrays associated with a primitive state token
 data MutableByteArray s = MutableByteArray (MutableByteArray# s)
                                         deriving( Typeable )
+
+instance NFData ByteArray where
+  rnf (ByteArray _) = ()
+
+instance NFData (MutableByteArray s) where
+  rnf (MutableByteArray _) = ()
 
 -- | Create a new mutable byte array of the specified size in bytes.
 newByteArray :: PrimMonad m => Int -> m (MutableByteArray (PrimState m))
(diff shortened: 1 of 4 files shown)


Data/Primitive/ByteArray.hs
```diff
@@ -98,6 +100,9 @@ data ByteArray = ByteArray ByteArray# deriving ( Typeable )
 data MutableByteArray s = MutableByteArray (MutableByteArray# s)
                                         deriving( Typeable )
 
+instance NFData ByteArray where
+  rnf = rwhnf
```
lehins, line 104: `rwnf` is only available since `deepseq-1.4.3.0`, which started to be shipped with gch-8.2.

I suggest a simple pattern match:
```haskell
instance NFData ByteArray where
  rnf (ByteArray _) = ()
```

I could also recommend adding one for `MutableByteArray`, which will work as intended since all elements must always be in normal form:

```haskell
instance NFData (MutableByteArray s) where
  rnf (MutableByteArray _) = ()
```

Same thing applies to `PrimArray` and `MutablePrimArray`

andrewthad: I agree. This also makes it much more clear that it is correct.

andrewthad:
This needs a changelog entry.

primitive.cabal
```diff
@@ -54,6 +54,7 @@ Library
         Data.Primitive.Internal.Operations
 
   Build-Depends: base >= 4.5 && < 4.15
+               , deepseq
```
chessai, line 57:
```suggestion
               , deepseq >= 1.4 && < 1.5
```

primitive.cabal
```diff
@@ -54,7 +54,7 @@ Library
         Data.Primitive.Internal.Operations
 
   Build-Depends: base >= 4.5 && < 4.15
-               , deepseq
+               , deepseq >= 1.4 && < 1.5
```
treeowl, line 57: Why do we need such tight bounds?

chessai: I think we could technically allow `>= 1.1 && < 1.5`, but I'm not sure what the point would be.

chessai: After the CI failure, I now see what the point would be.

primitive.cabal
```diff
@@ -54,6 +54,7 @@ Library
         Data.Primitive.Internal.Operations
 
   Build-Depends: base >= 4.5 && < 4.15
+               , deepseq >= 1.4 && < 1.5
```
chessai, line 57:
```suggestion
               , deepseq >= 1.1 && < 1.5
```

primitive.cabal
treeowl: Yeah, don't mess with build plans people may need unless you have a good reason.

andrewthad:
Squashed and merged. Thank you for this contribution. Also worthwhile would be adding NFData instances for `Array` and `SmallArray` if anyone wants to give that a shot.

Boarders:
I can give that a go too! Should be able to do it later today.

PR #272, opened 2020-06-05
Add NFData instances for Array and SmallArray

Boarders:
This adds `NFData` instances for `Array` and `SmallArray` as suggested by @andrewthad. I'm not too familiar with using CPP for supporting many versions so I'm happy to fix glaring errors.

### DIFF
--- a/Data/Primitive/Array.hs
+++ b/Data/Primitive/Array.hs
@@ -27,6 +27,7 @@ module Data.Primitive.Array (
   traverseArrayP
 ) where
 
+import Control.DeepSeq
 import Control.Monad.Primitive
 
 import GHC.Base  ( Int(..) )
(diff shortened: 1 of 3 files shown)


Data/Primitive/Array.hs
```diff
@@ -87,6 +89,18 @@ data Array a = Array
   { array# :: Array# a }
   deriving ( Typeable )
 
+#if MIN_VERSION_deepseq(1,4,3)
+instance NFData1 Array where
+  liftRnf r = Foldable.foldl' (\_ -> r) ()
+#endif
+
+instance NFData a => NFData (Array a) where
+#if MIN_VERSION_deepseq(1,4,3)
```
andrewthad, line 98: There not a good reason to switch out the implementation of `rnf` based on the availability of `liftRnf`. Better would be to just unconditionally do `Foldable.foldl' (\_ -> rnf) ()`, or if the code duplication concerns you, factor out `Foldable.foldl' (\_ -> rnf) ()` into a top-level `liftRnfArray` function and just call that in both places.

Boarders: Sounds good to me.

Issue #212, opened 2018-10-09
Bounds checking in array functions


osa1:
We discovered today that array functions do not do bounds checking, which was surprising to us. The expectation was that unless the function is prefixed with "unsafe" it's supposed to do bounds checking. 

After discovering this I checked the haddocks, but there's literally no mention to bounds checking in any of the array modules or at the top level. At the very least we should mention in bold letters that the array modules do not do bounds checking, and are completely unsafe.

For long-time primitive users and maintainers this is probably obvious, but it's far from being obvious from a user's point of view.

I realize that it's hard to change this behavior now, so perhaps we should just update the documentation. Ideally it'd be great to have `writeArray` that does bounds checking, and `unsafeWriteArray` that omits it though.

treeowl:
Functions in this package are generally *not* safe, and are typically named after the underlying primitives. Improving documentation is always helpful. I would oppose changing the semantics now, because packages using this one rely on the efficiency of unchecked operations. I'm not sure if the checked operations belong in this package or not.

cartazio:
i have to agree with David here, use Vector or something similar.

the cost/perf model for primitive needs to be brutally simple and
transparent. Any changes to that would otherwise impact performance /
semantics of eg vector (which has bounds checking / flags for it)

cartazio:
if you want safe arrays, use vector or array.

treeowl:
I *disagree* with Carter's suggestion. I think it's perfectly reasonable to
write checked operations for the arrays in this package. Not everyone wants
the stream fusion and slicing of vector, and not everyone wants the
flexible indexing of array. The question is whether those checked
operations belong in this package or in a separate (safe-primitive?) one. I
don't have a strong opinion either way.

cartazio:
we could add check ones, or an opt in checking flag, i'm merely opposed to
default built for current ops to do the checking ...

treeowl:
We could even do both, adding `safe` and `unsafe` variants, and making the
bare name depend on a flag. That should give maximal flexibility.

cartazio:
true. I'd rather keep primitive a tad simpler in that respect. (eg have a build flag for checked mode or something)

but for now, lets keep this as a backburner todo :)

cartazio:
for now lets approach this as "add a package flag for bounds checking operations "

chessai:
@osa1 this isn't precisely what you want (i think), but may be of interest to you: http://hackage.haskell.org/package/primitive-checked

[...]

PR #274, opened 2020-06-05
Add documentation

Boarders:
This adds documentation to some undocumented functions. It also adds notes that bounds (and other) checks are not performed (#212). In #212 it was mentioned that the intended audience already knows what they are doing but I think it is nevertheless useful to mentio and has no particular downsides. I may have gotten some of the checking notes wrong so let me know if so.

### DIFF
--- a/Data/Primitive/PrimArray.hs
+++ b/Data/Primitive/PrimArray.hs
@@ -217,9 +217,14 @@ instance (Show a, Prim a) => Show (PrimArray a) where
 die :: String -> String -> a
 die fun problem = error $ "Data.Primitive.PrimArray." ++ fun ++ ": " ++ problem
 
+-- | Create a 'PrimArray' from a list.
+--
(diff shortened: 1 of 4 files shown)


Data/Primitive/PrimArray.hs
```diff
@@ -480,6 +502,8 @@ unsafeThawPrimArray (PrimArray arr#)
   = primitive (\s# -> (# s#, MutablePrimArray (unsafeCoerce# arr#) #))
 
 -- | Read a primitive value from the primitive array.
+--
+-- /Note:/ this function does not\ do bounds checking.
```
andrewthad, line 506: What's this backslash doing here?

Boarders: Apologies, that's an accidental keyboard hit, let me fix it.

chessai:
Not sure how much of a visual difference this will make, but you left the colons in the italic blocks

Issue #211, opened 2018-08-30
Convenience classes


treeowl:
The state token type can be annoying to handle in constraints. I'd like something like

```haskell
class (PrimMonad m, PrimState m ~ s) => PrimMonadOn s m
instance (PrimMonad m, PrimState m ~ s) => PrimMonadOn s m

class (PrimBase m, PrimState m ~ s) => PrimBaseOn s m
instance (PrimBase m, PrimState m ~ s) => PrimBaseOn s m
```

I'm not fussy about the names.

ekmett:
I'd suggest `MonadPrim` and, less assuredly, something like `MonadPrimBase`.

That way you could use them in the style of the `mtl` classes, and never even notice the type families for most users.

cartazio:
that does sound like a reasonable bikeshed color Ed

chessai:
i wanted this, so for the time being i put it in this package: https://github.com/haskell-primitive/primitive-convenience

it's not on hackage. the name 'primitive-convenience' is also bad. i'd rather have this in primitive, and not a separate package, since it does make working with primmonad less annoying.

ekmett:
@chessai could you add MonadPrimBase?

cartazio:
Im halpybto add this too.  Though I may not get to it till after icfp next
month

chessai:
@ekmett done. Check master

cartazio:
@chessai https://github.com/haskell-primitive/primitive-convenience/commit/1fc04b3bd21c5a80766e2777a9dd8a11f63136c7 i'm looking at that ... are you sure those should all be monad PrimBase instances?

cartazio:
the only not ST/IO/STE primBase instance is IdentityT ...

chessai:
Ah, I made a mistake. I copy-pasted in my phone and changed all the
relevant identifiers.

https://github.com/haskell-primitive/primitive-convenience/commit/1816796ae3a69e2fc74f825b07de269bad1be109
fixes this.

ekmett:
class (PrimBase m, s ~ PrimState m) => MonadPrimBase s m | m -> s
    instance (PrimBase m, s ~ PrimState m) => MonadPrimBase s m 

using UndecidableInstances, etc. would be a more robust definition.

Then the user literally can't care about whether or not they are using fundeps and this class or the original class.

As it is you have to actually distinguish between having this convenience class, which is not implied by the usual constraint and having the original class.

hsenag:
[...]

PR #277, opened 2020-06-12
Add convenience classes MonadPrim and MonadPrimBase

andrewthad:


lehins:
@andrewthad Thank you for this PR, I've been wanting this for a while!!!!

andrewthad:
@lehins Would you mind reviewing this? Mostly to check if the documentation makes sense.

Control/Monad/Primitive.hs
```diff
@@ -225,6 +229,20 @@ instance PrimBase (L.ST s) where
...
+--   'PrimState' equality constraints less, by witnessing that
+--   @s ~ 'PrimState' m@.
+class (PrimMonad m, s ~ PrimState m) => MonadPrim s m | m -> s where
+instance (PrimMonad m, s ~ PrimState m) => MonadPrim s m
+
+-- | 'PrimBase''s state token type can be annoying to handle
+--   in constraints. This typeclass lets users (visually) notice
+--   'PrimState' equality constraints less, by witnessing that
+--   @s ~ 'PrimState' m@.
+class (PrimBase m, s ~ PrimState m) => MonadPrimBase s m | m -> s where
```
lehins, line 243: I think it would make sense to define this class in a similar fashion that `PrimBase` is defined:
```haskell
class (PrimBase m, MonadPrim s m) => MonadPrimBase s m
````

File: Control/Monad/Primitive.hs [hunk 2]
```diff
@@ -225,6 +229,20 @@ instance PrimBase (L.ST s) where
   {-# INLINE internal #-}
 #endif
 
+-- | 'PrimMonad''s state token type can be annoying to handle
+--   in constraints. This typeclass lets users (visually) notice
+--   'PrimState' equality constraints less, by witnessing that
+--   @s ~ 'PrimState' m@.
+class (PrimMonad m, s ~ PrimState m) => MonadPrim s m | m -> s where
```
lehins, line 236:
```suggestion
class (PrimMonad m, s ~ PrimState m) => MonadPrim s m | m -> s
```
Where is redundant

andrewthad:
@treeowl @ekmett @chessai @cartazio 

I haven't thought super hard about this, and you had all thought about it some in https://github.com/haskell/primitive/issues/211. Above, @lehins suggests that

    class (PrimBase m, s ~ PrimState m) => MonadPrimBase s m | m -> s where 

be changed to

    class (PrimBase m, MonadPrim s m) => MonadPrimBase s m

Does anyone see any difference between these two approaches?

emilypi:
Wonderful, thank you 👏

cartazio:
I’m not sure if they propagate quite the same way. At least in older ghcs they work differently. I can’t recall at the moment what ghc version starts desugaring them the same way.  Or if that change is just a paper by Richard.  I feel like @RyanGlScott or @ekmett  Or perhaps @hvr will have some clarity.  

For more complicated fundep relations or older ghc theres definitely a difference.

emilypi:
Maybe there is old lore I don't undersand about this, but testing locally seems to have to same results (modulo error messages being slightly different): 

```haskell
П> class A (m :: * -> *)
П> instance A []
П> type family B (m :: * -> *) :: *
П> type instance B [] = Int
П> instance C Int []
П> instance C Char []
П> class D (m :: * -> *)
П> class (D m, C s m) => E s m
П> instance D []
П> instance E Char []

<interactive>:22:10: error:
    • Couldn't match type ‘Int’ with ‘Char’
        arising from a functional dependency between:
          constraint ‘C Char []’
            arising from the superclasses of an instance declaration
          instance ‘C Int []’ at <interactive>:17:10-17
    • In the instance declaration for ‘E Char []’
П> instance E Int []
П>
```

vs 

```haskell
П> class A (m :: * -> *)
П> instance A []
П> type family B (m :: * -> *) :: *
П> type instance B [] = Int
П> class C (m :: * -> *)
П> instance C []
П> class (C m, s ~ B m) => D s m | m -> s
П> instance D Int []
П> instance D Char []

<interactive>:10:10: error:
    Functional dependencies conflict between instance declarations:
      instance [safe] D Int [] -- Defined at <interactive>:10:10
      instance D Char [] -- Defined at <interactive>:11:10
```

RyanGlScott:
I can't think of any difference between these two versions of `MonadPrimBase` with respect to type inference. There is a slight difference in the underlying Core, however. This version:

> `class (PrimBase m, s ~ PrimState m) => MonadPrimBase s m`

Will compile down to a dictionary with two fields of types `PrimBase m` and `s ~ PrimState m`, respectively. This version:

> `class (PrimBase m, MonadPrim s m) => MonadPrimBase s m`

Will also compile down to a two-field dictionary, but this time, the `s ~ PrimState m` field is tucked underneath a `MonadPrim s m` dictionary. A consequence of this is that if you need to use `MonadPrimBase` to deduce that `s ~ PrimState m`, as in the following example:

```hs
f :: MonadPrimBase s m => proxy m -> s :~: PrimState m
f _ = Refl
```

Then with the first encoding of `MonadPrimBase`, the resulting Core needs to pattern-match on a single dictionary to extract out the evidence for `s ~ PrimState m`. With the second version, however, the resulting Core needs to pattern-match on two dictionaries.

Is this difference actually significant enough to matter in practice? I dunno. I'll let you be the judge of that :)

emilypi:
@RyanGlScott thanks, that makes a lot of sense.

chessai:
The only argument I can think of for why we should include the explicit fundep (i.e. not just letting it propagate) is for slightly better documentation (your eyes immediately see the fundep). But, this is so minour I'm not sure it matters at all. I think in practise that this will never cause UX issues?

chessai:
A good argument for why it should be defined using propagation is uniformity, which I think is stronger than the minour detail of eye-catching fundeps. So I am +1 on defining it how @lehins mentioned.

lehins:
I suggested an alternative definition because I though it might feel cleaner and avoid an extra fun dep, but that this is of course a pretty subjective suggestion. Whichever approach @andrewthad will choose will not matter to me I like this PR in either form ;)

treeowl:
I don't see the point of the fundep. Franky, its redundancy made me do a double take. Can we get rid of that?

treeowl:
@andrewthad, I'm sorry I'm late to this party, but I want to express my support for this class. I personally wish it were the fundamental one, but that ship has sailed. I'm not a big fan of the single-parameter-with-associated-type style, and I'm glad you've given us an alternative.

treeowl:
Ahaha! Just noticed I was the one who opened the issue way back when. Forgot the whole thing.

chessai:
see #288

cartazio:
One option to put them on equal footing is have the default impl for the
new and old classes be each other?  I’ve not thought through the
implications, but that’s one option to consider.

PR #285, opened 2020-06-12
Add freeze shims for ByteArray and PrimArray

andrewthad:


andrewthad:
I forgot the changelog entry, but that was added in 9e895b93a2ef8735082b1717e387cf0f96097c72

PR #288, opened 2020-06-21
remove fundep on MonadPrim

chessai:
cc @emilypi @treeowl 

I think we just auto-fundep'd based on MTL. it is indeed unnecessary.


emilypi:
👍  from me

treeowl:
I'm very confused about https://github.com/haskell/primitive/issues/211#issuecomment-512601248 which seems to be where the fundeps came from. I have no idea what @ekmett was talking about there.

PR #300, opened 2020-12-05
Update QuickCheck upper bounds

lehins:


andrewthad:
Thanks!

lehins:
Very nice. @andrewthad any chance of a hackage revision, just so we don't have to turn off the test suite on stackage?

andrewthad:
Done.

lehins:
Awesome. Thank you!

PR #309, opened 2021-07-01
Zero-pad words encoded in ByteArray Show instance

andrewthad:
Previously, the Show instance for ByteArray would produce results like [0xb, 0x1f, 0x5]. Now, it produces [0x0b, 0x1f, 0x05].

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,9 +2,8 @@
 
   * Add `thawByteArray` and `thawPrimArray`.
 
-  * Changed the `Show` instance of `ByteArray` so that all 8-bit words
-    are rendered as two digits. For example, display `0x0D` instead of `0xD`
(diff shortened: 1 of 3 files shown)


changelog.md
```diff
@@ -2,6 +2,10 @@
 
   * Add `thawByteArray` and `thawPrimArray`.
 
+  * Changed the `Show` instance of `ByteArray` so that all 8-bit words
+    are rendered as two digits. For example, display `0x0D` instead of `0xD`
+    for as an element.
```
konsumlamm, line 7:
```suggestion
  * Changed the `Show` instance of `ByteArray`, so that all 8-bit words
    are rendered as two digits. For example, display `0x0D` instead of `0xD`.
```

PR #318, opened 2021-08-29
Fix bugs in `*>`

konsumlamm:
The implementations of `*>` for `Array` and `SmallArray` were buggy:

* `Array`: The bug was caused by not calling `go (i + 1)`.
* `SmallArray`: The bug was caused by using `0` instead of `i * lb` as offset when calling `copySmallArray` in the implementation of `*>`.

I also did some minor formatting changes.

Btw, is there any particular reason for using `fix` instead of defining a recursive `go` function directly?

### DIFF
--- a/Data/Primitive/Array.hs
+++ b/Data/Primitive/Array.hs
@@ -672,33 +672,36 @@ instance Functor Array where
 
 instance Applicative Array where
   pure x = runArray $ newArray 1 x
-  ab <*> a = createArray (szab*sza) (die "<*>" "impossible") $ \mb ->
+
+  ab <*> a = createArray (szab * sza) (die "<*>" "impossible") $ \mb ->
     let go1 i = when (i < szab) $
             do
               f <- indexArrayM ab i
-              go2 (i*sza) f 0
-              go1 (i+1)
+              go2 (i * sza) f 0
+              go1 (i + 1)
         go2 off f j = when (j < sza) $
             do
               x <- indexArrayM a j
               writeArray mb (off + j) (f x)
               go2 off f (j + 1)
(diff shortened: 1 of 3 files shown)


treeowl:
There's so much lint noise (for whatever reason) that I can hardly see what's actually been changed. Not your fault, but really annoying. Could you please add test properties to demonstrate the fix?

konsumlamm:
I added property tests for `*>` and `<*`, which just compare the result with the respective list implementation.
An example of what failed before (with an exception) for both `Array` and `SmallArray` (imagine the lists to be arrays):
```hs
[1, 2] *> [3, 4, 5]
```

konsumlamm:
@treeowl ping

treeowl:
I don't seem to have commit bits for this package anymore. @andrewthad ?

chessai (approved):

chessai:
> I don't seem to have commit bits for this package anymore. @andrewthad ?

"treeowl has been added as a collaborator with maintain permissions on the repository."

Just added you

chessai:
@konsumlamm thanks for the patch!

andrewthad:
For what it's worth, I had never never removed you, so I don't know how that happened.

andrewthad:
I'll get out a release for this soon in the next few days since this is a serious issue.

treeowl:
Very strange. I know I went dormant for a while; I figured you decided I
was gone.

chessai:
I did not remove you either.

treeowl:
Curiouser and curiouser. Well, it's fixed now.

chessai:
I did not remove you either.

PR #319, opened 2021-08-31
Export some useful functions

### LINKED ISSUE #185: some useful functions remain unexported
Data.Primitive.Array.createArray & Data.Primitive.SmallArray.createSmallArray

seem to me like both are totally safe ways of constructing arrays. Is there a reason they are not exported?

konsumlamm:
Closes #185.

Export `runByteArray`, `runPrimArray`, `emptyByteArray`, `emptyPrimArray`, `emptyArray` and `emptySmallArray`. I also added doc comments.

EDIT: Also export `createArray` and `createSmallArray`.

### DIFF
--- a/Data/Primitive/ByteArray.hs
+++ b/Data/Primitive/ByteArray.hs
@@ -32,6 +32,7 @@ module Data.Primitive.ByteArray (
   readByteArray, writeByteArray, indexByteArray,
 
   -- * Constructing
+  emptyByteArray,
   byteArrayFromList, byteArrayFromListN,
 
   -- * Folding
@@ -41,7 +42,7 @@ module Data.Primitive.ByteArray (
   compareByteArrays,
 
   -- * Freezing and thawing
-  freezeByteArray, thawByteArray,
+  freezeByteArray, thawByteArray, runByteArray,
   unsafeFreezeByteArray, unsafeThawByteArray,
 
   -- * Block operations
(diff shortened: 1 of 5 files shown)


Data/Primitive/ByteArray.hs
```diff
@@ -696,7 +697,9 @@ calcLength :: [ByteArray] -> Int -> Int
 calcLength [] !n = n
 calcLength (x : xs) !n = calcLength xs (sizeofByteArray x + n)
 
+-- | The empty 'ByteArray'.
 emptyByteArray :: ByteArray
+{-# NOINLINE emptyByteArray #-}
```
konsumlamm, line 702: I added a `NOINLINE` pragma for `emptyByteArray`, since all the other empty arrays also have that (I assume this was just an oversight).

treeowl: Yes, just an oversight. We don't really want to allocate extra empty arrays for no reason.

treeowl:
I just looked again... I think I was wrong about `createArray`. Sorry! Just be sure the documentation explains the difference.

konsumlamm:
Done.

konsumlamm:
@treeowl now that you're a maintainer, is there anything stopping this from getting merged?

konsumlamm:
> Yes, the documentation for `createArray` is insufficient. Please explain what makes it different from `runArray`.

@treeowl can you elaborate what you mean? The doc comment says what it's doing step by step and even shows how it's (conceptually) defined in terms of `runArray`. Do you have a suggestion how to make the doc comment more clear?

treeowl:
Sorry, I should've put that differently. What I mean is that the documentation should point out that if the size is 0, then the result will be the canonical empty array, rather than a fresh copy thereof.

treeowl:
Thanks!

PR #321, opened 2021-09-14
Drop support for GHC < 7.10

### LINKED ISSUE #292: drop support for older GHCs in 0.8 release
Recently, both vector and random have dropped support for anything pre-7.10. I think this is reasonable, since 7.0-7.8 are really old, and impose some annoying maintenance burdens.

konsumlamm:
Closes #292.

I bumped the minimum `base` version to 4.8 and the minimum version of `transformers` to 0.4.2 (since that is the first version supporting `base >= 4.8`). Most notably, this removes the need for `HAVE_SMALL_ARRAY` (and the implementation of `Data.Primitive.SmallArray` in terms of `Data.Primitive.Array` in general).


treeowl:
Wow, that's a lot of cruft you ripped out. Nice work.

andrewthad:
I haven't audited this carefully, but if the test suite is passing, that gives me considerable confidence that it didn't break anything. Great work!

PR #322, opened 2021-09-17
Test GHC 7.10 in CI

konsumlamm:


konsumlamm:
It seems the issue was fixed in https://github.com/haskell/actions/pull/83, so I'll try again.

konsumlamm:
The test suite passes now (the CI failure is caused by a timeout).

konsumlamm:
@andrewthad could you please take another look?

konsumlamm:
@andrewthad ping

The CI problem seems to be the same as in https://github.com/haskell/primitive/pull/326#issuecomment-1000393728.

andrewthad:
This is just a guess, but try changing this:

```
    - name: Build
      run: cabal build
```

to this:

```
    - name: Build
      run: cabal build --enable-tests
```

konsumlamm:
@andrewthad that doesn't change anything, although it shows that the problem is not with building the tests suite (see https://github.com/haskell/primitive/runs/4622761485).

Using `cabal configure --enable-tests` seems to work (although for some reason the last task can't find any runner).

I've marked this PR as a draft, because the CI currently fails for GHC 7.10 (see #329). Once that is merged, I'll readd 7.10 and while we're at it, we can also add 9.2 (although technically it's already tested, since `latest` is 9.2.1 currently).

andrewthad:
Closing since GHC 7.10 will no longer be supported.

PR #326, opened 2021-12-23
Add/update definitions for `stimes`

konsumlamm:
* add `stimes` definition for `Array` & `SmallArray`
* change `stimes` definition for `ByteArray` (and thus also for `ByteArray`) to error for negative multipliers and not check for overflow (no other `primitive` functions check for overflows, so this seems pointless)
* add property tests for `stimes` & `semigroupLaws`


konsumlamm:
I really don't know what's going on with the CI. Running the tests locally works fine and the macOS tests seem to work when the `cabal.project` file contains a `tests: True` line.

andrewthad:
Thanks. All of this looks good. Merging.

PR #329, opened 2021-12-24
Fix `semigroupLaws` tests

konsumlamm:
Followup to #326.

The test suite would previously fail on 7.10, because I forgot to put the `semigroupLaws` test inside a `#if` block (sorry).


andrewthad:
Can you confirm that this is good to be merged? I'm can't tell based on CI output, but I'll take your word if you say it works.

PR #332, opened 2022-01-12
Add Lift instances

### LINKED ISSUE #331: Add template-haskell Lift instances

treeowl:
Closes #331

### DIFF
--- a/Data/Primitive/Array.hs
+++ b/Data/Primitive/Array.hs
@@ -75,30 +75,33 @@ data Array a = Array
 
 instance Lift a => Lift (Array a) where
 #if MIN_VERSION_template_haskell(2,16,0)
-  liftTyped ary
-    | len == 0 = [|| Array (emptyArray# (##)) ||]
-    | len == 1 = [|| pure frst ||]
-    | otherwise = [|| unsafeArrayFromListN' len lst ||]
+  liftTyped ary = case lst of
+    [] -> [|| Array (emptyArray# (##)) ||]
+    [x] -> [|| pure x ||]
+    x : xs -> [|| unsafeArrayFromListN' len x xs ||]
 #else
-  lift ary = [| unsafeArrayFromListN' len lst |]
+  lift ary = case lst of
+    [] -> [| Array (emptyArray# (##)) |]
+    [x] -> [| pure x |]
+    x : xs -> [| unsafeArrayFromListN' len x xs |]
 #endif
     where
       len = length ary
       lst = toList ary
-      frst = ary `indexArray` 0
 
--- | Strictly create an array from a list of a known length. If the length
+-- | Strictly create an array from a nonempty list (represented as
+-- a first element and a list of the rest) of a known length. If the length
 -- of the list does not match the given length, this makes demons fly
 -- out of your nose. We use it in the 'Lift' instance. If you edit the
 -- splice and break it, you get to keep both pieces.
-unsafeArrayFromListN' :: Int -> [a] -> Array a
-unsafeArrayFromListN' n l =
-  createArray n (die "fromListN" "uninitialized element") $ \sma ->
+unsafeArrayFromListN' :: Int -> a -> [a] -> Array a
+unsafeArrayFromListN' n y ys =
+  createArray n y $ \ma ->
     let go !_ix [] = return ()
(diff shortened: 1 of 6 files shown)


Data/Primitive/ByteArray.hs
```diff
@@ -413,6 +457,23 @@ copyByteArrayToPtr (Ptr dst#) (ByteArray src#) soff sz
...
+  -> Int -- ^ offset into destination array, interpreted as elements of type @a@
+  -> Ptr a -- ^ source
+  -> Int -- ^ number of elements to copy
+  -> m ()
+{-# INLINE copyPtrToByteArray #-}
+copyPtrToByteArray (MutableByteArray dst#) soff (Ptr src#) sz
+  = primitive_ (copyAddrToByteArray# src# dst# (unI# soff *# siz# ) (unI# sz))
+  where
+  siz# = sizeOf# (undefined :: a)
+
```
treeowl, line 476: Surely we want to export `copyPtrToByteArray`, whether with this argument order or a different one, but that's for another PR.

Data/Primitive/Array.hs
```diff
@@ -65,12 +66,40 @@ import qualified Text.ParserCombinators.ReadPrec as RdPrc
...
+    where
+      len = length ary
+      lst = toList ary
+      frst = ary `indexArray` 0
+
+-- | Strictly create an array from a list of a known length. If the length
+-- of the list does not match the given length, this makes demons fly
+-- out of your nose. We use it in the 'Lift' instance. If you edit the
+-- splice and break it, you get to keep both pieces.
+unsafeArrayFromListN' :: Int -> [a] -> Array a
```
treeowl, line 94: What would happen if we wrote this as a fold? Would GHC be able to make good decisions about when to build it as a list and convert it to an array vs. building it directly as an array (poke poke poke)?

Data/Primitive/ByteArray.hs
phadej: There is https://hackage.haskell.org/package/primitive-0.7.3.0/docs/Data-Primitive-Ptr.html#v:copyPtrToMutablePrimArray

Is it the same?

treeowl: Aha! I guess we just need to move its definition.

phadej: :+1:

phadej: Inm the name of consistency, should copyPtrToMutablePrimArray be exported from `PrimArray` module as well?

Data/Primitive/Array.hs
phadej: I can check this tomorrow.

Data/Primitive/ByteArray.hs
treeowl: Yes, I think so.

treeowl: Added a commit to export `copyPtrToMutablePrimArray` from `PrimArray`.

Data/Primitive/Array.hs
treeowl: If I make the fold `INLINABLE`, it doesn't actually seem to do anything. If I make it `INLINE`, I get pokes up to around 32 elements, which suggests some judgement has been used. Unfortunately, going bigger seems to lead to some useless code duplication. So we might be better off just choosing our own cutoff, whatever that should be (currently set to 1, essentially). Another reasonable option would be to just leave it as I have it right now and push GHC for array literals or something.

Data/Primitive/ByteArray.hs
```diff
@@ -118,7 +118,7 @@ instance TH.Lift ByteArray where
 #if __GLASGOW_HASKELL__ >= 802
       pinned = isByteArrayPinned ba
 #else
-      pinned = False
+      pinned = True
```
phadej, line 121: Uh.

> This function is only available when compiling with GHC 8.2 or newer.

on `isByteArrayPinned` is almost impossible to spot.

`isByteArrayPinned#` in `base` doesn't have `@since` annotations either.

phadej (approved):
This looks great!

treeowl:
Thanks, @phadej! I'll merge.

PR #339, opened 2022-05-03
Drop support for GHC 7.10

### LINKED ISSUE #337: `primitive` fails to compile for 7.10
#332 introduced `Lfit` instances. However, this uses the `TemplateHaskellQuotes` extension, which is only available since GHC 8.0. There are several options how to resolve this:

* drop 7.10 support
* use the `TemplateHaskell` extension for 7.10
* don't define the `Lift` instances for 7.10

EDIT: The `Lift` instances also use `emptyArray#`/`emptySmallArray#`, which isn't available either for GHC 7.10.

konsumlamm:
Resolves #337.

This also gets rid of a bunch of CPP. I bumped the version bounds to be appropriate for base >= 4.9.

### DIFF
--- a/Data/Primitive/Array.hs
+++ b/Data/Primitive/Array.hs
@@ -48,15 +48,11 @@ import Control.Monad.Fix
 import qualified Data.Foldable as Foldable
 import Control.Monad.Zip
 import Data.Foldable (Foldable(..), toList)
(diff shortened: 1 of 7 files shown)


andrewthad:
Looks good. Merging. And thanks for doing this!

PR #340, opened 2022-05-19
GHC 9.4 support

parsonsmatt:
The current release on Hackage fails to build with 9.4, but the version on `master` works fine with `--allow-newer` (with a few warnings about some new syntax).

Omitting `--allow-newer` gives us an error with QuickCheck and `splitmix`, but those are also fine for now.

### DIFF
--- a/primitive.cabal
+++ b/primitive.cabal
@@ -48,7 +48,7 @@ Library
   Other-Modules:
         Data.Primitive.Internal.Operations
 
-  Build-Depends: base >= 4.9 && < 4.17
+  Build-Depends: base >= 4.9 && < 4.18
                , deepseq >= 1.1 && < 1.5
                , transformers >= 0.5 && < 0.7
                , template-haskell >= 2.11


parsonsmatt:
A release on this would be great, too 😅

andrewthad:
Merged. Do you need an actual release, or would a hackage revision be alright?

parsonsmatt:
Hackage revision would be fine, thank you!

parsonsmatt:
Since it is only a bounds change, a revision should be fine. I'd need to do `cabal update` but that's not a huge deal. Am I missing something?

Bodigrim:
You seem to assume that your change is the only one atop of the released `primitive-0.7.3.0`. This is not so. 
@andrewthad please make a new release.

andrewthad:
@Bodigrim Good catch. I've prepared a release candidate at https://github.com/haskell/primitive/pull/341. It just bumps the version and updates the changelog.

@parsonsmatt I don't actually have a build of GHC 9.4 available locally. Have you confirmed that the test suite passes (not just builds)? If you have, I can release.

andrewthad:
Released!

PR #343, opened 2022-06-02
Add PrimRef

oberblastmeister:
I copied it from `prim-ref`


konsumlamm:
A few comments:
* The module should also be exported, otherwise there isn't much point in adding it.
* I'm not sure why/if `FrozenPrimRef` is needed, since a `FrozenPrimRef a` is basically just an `a`.
* I think the implementation could be simplified by using a `PrimArray` under the hood (Rather than a `ByteArray`). That way we don't have to duplicate all the `sizeOf` stuff.
* About the name, I think the best candidates are `PrimRef` (to be consistent with `IORef` and `STRef`) and `PrimVar` (to be consistent with `MutVar`). I tend to prefer the latter, since  `MutVar` is also defined in `primitive`.

oberblastmeister:
I think `PrimRef` is better, because the `Var` names make me think of concurrency. But of course we have `MutVar` already.

konsumlamm:
I think `PVar` would be a bad name, it's not at all obvious what the `P` stands for (yes, the same thing can be said about `MVar` and `TVar`, but I don't think those names are great either, although I don't have a better name for them). `PrimVar` on the other hand implies that it uses `Prim` / is based on `PrimArray` (which isn't named `PArray` after all) and it isn't much longer.

lehins:
@konsumlamm No argument from my side. `PrimVar` is good. @andrewthad what's your take on it

andrewthad:
Let’s go with PrimVar since it has a since symmetry with MutVar. The frozen variant seems useless though. I think that should be eliminated.

oberblastmeister:
I wonder if we could call the atomic functions `casIntVar` instead of `casIntPrimVar`. The latter seems a bit verbose.

Data/Primitive/PrimRef.hs
```diff
@@ -0,0 +1,146 @@
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE MagicHash #-}
+{-# LANGUAGE RoleAnnotations #-}
+{-# LANGUAGE UnboxedTuples #-}
+{-# LANGUAGE Unsafe #-}
+
+-- | Note: Edward Kmett wrote everything in this module.
```
konsumlamm, line 7:
```suggestion
-- | Unboxed mutable variables.
--
-- Note: Edward Kmett wrote everything in this module. 
```
Should we keep the note about Kmett here? The code is pretty trivial, it's just wrapping `MutablePrimArray` functions and some atomic functions from `GHC.Exts`. The original also used `MutableByteArray`.

Data/Primitive/PrimRef.hs
```diff
@@ -0,0 +1,146 @@
...
+import Control.Monad.Primitive
+import Data.Primitive
+import GHC.Exts
+import GHC.Ptr (castPtr)
+
+--------------------------------------------------------------------------------
+-- * Primitive References
+--------------------------------------------------------------------------------
+
+newtype PrimVar s a = PrimVar (MutablePrimArray s a)
```
konsumlamm, line 39:
```suggestion
-- | A 'PrimVar' behaves like a single-element mutable primtiive array.
newtype PrimVar s a = PrimVar (MutablePrimArray s a)
```

Data/Primitive/PrimRef.hs
```diff
@@ -0,0 +1,146 @@
...
+
+type role PrimVar nominal nominal
+
```
konsumlamm, line 42:
```suggestion
```
I'm not sure this is necessary, at least the other types in this package don't have any role annotations. The `s` seems to be nominal anyway, while `a` is phantom by default.

Data/Primitive/PrimRef.hs
```diff
@@ -0,0 +1,146 @@
...
+-- | Create a primitive reference.
```
konsumlamm, line 43:
```suggestion
-- | Create a 'PrimVar'.
```
In general, I'd replace `primitive reference` by `'PrimVar'`.

Data/Primitive/PrimRef.hs
```diff
@@ -0,0 +1,146 @@
...
+
+-- | Create a pinned primitive reference with the appropriate alignment for its contents.
+newAlignedPinnedPrimVar :: (PrimMonad m, Prim a) => a -> m (PrimVar (PrimState m) a)
+newAlignedPinnedPrimVar a = do
+  m <- newAlignedPinnedPrimArray 1
+  writePrimArray m 0 a
+  return (PrimVar m)
+{-# INLINE newAlignedPinnedPrimVar #-}
+
+-- | Read a primitive value from the reference
```
konsumlamm, line 67:
```suggestion
-- | Read a value from the 'PrimVar'.
```

Data/Primitive/PrimRef.hs
```diff
@@ -0,0 +1,146 @@
...
+readPrimVar :: (PrimMonad m, Prim a) => PrimVar (PrimState m) a -> m a
+readPrimVar (PrimVar m) = readPrimArray m 0
+{-# INLINE readPrimVar #-}
+
+-- | Write a primitive value to the reference
```
konsumlamm, line 72:
```suggestion
-- | Write a value to the 'PrimVar'.
```

Data/Primitive/PrimRef.hs
```diff
@@ -0,0 +1,146 @@
...
+  PrimVar m == PrimVar n = sameMutablePrimArray m n
+  {-# INLINE (==) #-}
+
+-- | Yield a pointer to the data of a 'PrimVar'. This operation is only safe on pinned byte arrays allocated by
+-- 'newPinnedPrimVar' or 'newAlignedPinnedPrimVar'.
+primVarContents :: PrimVar s a -> Ptr a
+primVarContents (PrimVar m) = castPtr $ mutablePrimArrayContents m
+{-# INLINE primVarContents #-}
+
+primVarToMutablePrimArray :: PrimVar s a -> MutablePrimArray s a
```
konsumlamm, line 87:
```suggestion
-- | Return the underlying 'MutablePrimArray'.
primVarToMutablePrimArray :: PrimVar s a -> MutablePrimArray s a
```

Data/Primitive/PrimRef.hs
```diff
@@ -0,0 +1,146 @@
...
+primVarToMutablePrimArray (PrimVar m) = m
+{-# INLINE primVarToMutablePrimArray #-}
+
+--------------------------------------------------------------------------------
+-- * Atomic Operations
+--------------------------------------------------------------------------------
+
+-- | Given a primitive reference, the expected old value, and the new value, perform an atomic compare and swap i.e. write the new value if the current value matches the provided old value. Returns the value of the element before the operation. Implies a full memory barrier.
```
konsumlamm, line 95:
```suggestion
-- | Given a primitive reference, the expected old value, and the new value, perform an atomic compare and swap i.e. write the new value if the current value matches the provided old value.
-- Returns the value of the element before the operation. Implies a full memory barrier.
```

Data/Primitive/PrimRef.hs
```diff
@@ -0,0 +1,146 @@
...
+  (# s', result #) -> (# s', I# result #)
+{-# INLINE fetchOrInt #-}
+
+-- | Given a reference, and a value to bitwise xor, atomically xor the value with the element. Returns the value of the element before the operation. Implies a full memory barrier.
+fetchXorInt :: PrimMonad m => PrimVar (PrimState m) Int -> Int -> m Int
+fetchXorInt (PrimVar (MutablePrimArray m)) (I# x) = primitive $ \s -> case fetchXorIntArray# m 0# x s of
+  (# s', result #) -> (# s', I# result #)
+{-# INLINE fetchXorInt #-}
+
+-- | Given a reference, read an element. Implies a full memory barrier.
```
konsumlamm, line 137:
```suggestion
-- | Given a reference, atomically read an element. Implies a full memory barrier.
```

Data/Primitive/PrimRef.hs
```diff
@@ -0,0 +1,146 @@
...
+atomicReadInt :: PrimMonad m => PrimVar (PrimState m) Int -> m Int
+atomicReadInt (PrimVar (MutablePrimArray m)) = primitive $ \s -> case atomicReadIntArray# m 0# s of
+  (# s', result #) -> (# s', I# result #)
+{-# INLINE atomicReadInt #-}
+
+-- | Given a reference, write an element. Implies a full memory barrier.
```
konsumlamm, line 143:
```suggestion
-- | Given a reference, atomically write an element. Implies a full memory barrier.
```

Data/Primitive/PrimRef.hs
```diff
@@ -0,0 +1,146 @@
...
+  -- * Primitive References
+    PrimVar(..)
+  , newPrimVar
+  , newPinnedPrimVar
+  , newAlignedPinnedPrimVar
+  , readPrimVar
+  , writePrimVar
+  , primVarContents
+  , primVarToMutablePrimArray
+  , casInt
```
konsumlamm, line 19:
```suggestion
  -- * Atomic Operations
  -- $atomic
  , casInt
```

Data/Primitive/PrimRef.hs
```diff
@@ -0,0 +1,146 @@
...
+
+import Control.Monad.Primitive
+import Data.Primitive
+import GHC.Exts
+import GHC.Ptr (castPtr)
+
+--------------------------------------------------------------------------------
+-- * Primitive References
+--------------------------------------------------------------------------------
+
```
konsumlamm, line 38:
```suggestion

```

Data/Primitive/PrimRef.hs
```diff
@@ -0,0 +1,146 @@
...
+primVarContents (PrimVar m) = castPtr $ mutablePrimArrayContents m
+{-# INLINE primVarContents #-}
+
+primVarToMutablePrimArray :: PrimVar s a -> MutablePrimArray s a
+primVarToMutablePrimArray (PrimVar m) = m
+{-# INLINE primVarToMutablePrimArray #-}
+
+--------------------------------------------------------------------------------
+-- * Atomic Operations
+--------------------------------------------------------------------------------
```
konsumlamm, line 93:
```suggestion
-- $atomic
-- Atomic operations on `PrimVar s Int`. They imply a full memory barrier.
```
I'd remove the `Implies a full memory barrier.` part from the doc comments and instead just mention it here once.

konsumlamm (commented):
Looks good to me (once my comments are addressed).

konsumlamm:
> I wonder if we could call the atomic functions `casIntVar` instead of `casIntPrimVar`. The latter seems a bit verbose.

I think `casInt` (as it is currently) is fine.

konsumlamm:
I'd also like to have `modifyPrimVar` (analogous to `modifyMutVar`), but that can also be done in a separate PR.

Data/Primitive/PrimRef.hs
```diff
@@ -0,0 +1,146 @@
...
+writePrimVar (PrimVar m) a = writePrimArray m 0 a
+{-# INLINE writePrimVar #-}
+
+instance Eq (PrimVar s a) where
+  PrimVar m == PrimVar n = sameMutablePrimArray m n
+  {-# INLINE (==) #-}
+
+-- | Yield a pointer to the data of a 'PrimVar'. This operation is only safe on pinned byte arrays allocated by
+-- 'newPinnedPrimVar' or 'newAlignedPinnedPrimVar'.
+primVarContents :: PrimVar s a -> Ptr a
```
lehins, line 83: I'd recommend turning this function into a safe one by checking `isByteArrayPinned#` and returning `Maybe` instead.
```suggestion
primVarContents :: PrimVar s a -> Maybe (Ptr a)
```

lehins (approved):
> I think casInt (as it is currently) is fine.

I respectfully disagree. Even for consistency sake I'd suggest adding `PrimVar` suffix to all atomic operations. 

Another very useful function would be a `atomicModifyIntPrimVar`, which can be implemented in terms of `casIntPrimVar`. Feel free to copy paste this implementation: https://hackage.haskell.org/package/pvar-1.0.0.0/docs/src/Data.Primitive.PVar.Internal.html#atomicModifyIntArray%23

Data/Primitive/PrimRef.hs
oberblastmeister: Shouldn't the other types have role annotations? Doing this makes it safe, `Prim` cannot read out of bounds because of a type change.

Data/Primitive/PrimRef.hs
oberblastmeister: This is not really consistent with the rest of the library.

Data/Primitive/PrimRef.hs
oberblastmeister: This would be nice but is not consistent

Data/Primitive/PrimRef.hs
konsumlamm: Hmm, that's a good point. So representational should be fine for `a`, but we can also make it nominal for now and then relax it later.

Data/Primitive/PrimRef.hs
konsumlamm: It's consistent with the documentation of `Data.Primitive.MutVar`. At least it should talk about primitive *references*, since we decided to call the type `PrimVar`. Besides, I don't think what the rest of the library does is really important for this kind of things...

primitive.cabal
```diff
@@ -44,6 +44,7 @@ Library
         Data.Primitive.Ptr
         Data.Primitive.MutVar
         Data.Primitive.MVar
+        Data.Primitive.PrimRef
```
konsumlamm, line 47:
```suggestion
        Data.Primitive.PrimVar
```
I just noticed that the file is still called `PrimRef.hs`.

Data/Primitive/PrimRef.hs
oberblastmeister: I don't think representational would be fine, because in theory you could still give a different implementation of `Prim`.

oberblastmeister: We could also have a `unsafeCoercePrimVar` if a coercion is needed.

primitive.cabal
konsumlamm: @oberblastmeister you also need to rename the file.

oberblastmeister:
@andrewthad Anything else I should do here? I can add some more functions later

konsumlamm:
> Anything else I should do here?

I still have some unresolved comments, see above.

andrewthad:
Sorry, I've not reviewed this yet. I'll review this Friday.

Data/Primitive/PrimRef.hs
andrewthad: Agreed. That's inconsistent with several other modules in this library. Let's leave this without the `Maybe`, and a separate function could be added in another PR if someone needs that.

Data/Primitive/PrimRef.hs
andrewthad: Having the role be explicit is important. The inferred role (phantom) allows user to coerce a `PrimVar s Word8` to a `PrimVar s Word64`, which is unsound. `PrimArray` really should have a role annotation as well.

Data/Primitive/PrimRef.hs
andrewthad: I've rewritten this header in a PR against this branch. I think it's good to mention the original author.

andrewthad:
@oberblastmeister I've opened a PR against your branch with some documentation changes. Everything looks good to me after these changes are introduced.

@konsumlamm I've incorporated many of your suggestions into that PR.

oberblastmeister:
@andrewthad I've merged your changes.

Data/Primitive/PrimVar.hs
```diff
@@ -0,0 +1,159 @@
...
+  (# s', result #) -> (# s', I# result #)
+{-# INLINE fetchSubInt #-}
+
+-- | Given a reference, and a value to bitwise and, atomically and the value with the element. Returns the value of the element before the operation.
+fetchAndInt :: PrimMonad m => PrimVar (PrimState m) Int -> Int -> m Int
+fetchAndInt (PrimVar (MutablePrimArray m)) (I# x) = primitive $ \s -> case fetchAndIntArray# m 0# x s of
+  (# s', result #) -> (# s', I# result #)
+{-# INLINE fetchAndInt #-}
+
+-- | Given a reference, and a value to bitwise nand, atomically nand the value with the element. Returns the value of the element before the operation. Implies a full memory barrier.
```
konsumlamm, line 132:
```suggestion
-- | Given a reference, and a value to bitwise nand, atomically nand the value with the element. Returns the value of the element before the operation.
```

konsumlamm (changes requested):

andrewthad:
Squashed and merged. Thanks to everyone who made this happen!

PR #346, opened 2022-07-06
make PrimArray nominal

oberblastmeister:


andrewthad:
This looks good. Merging.

chessai:
This needs a changelog entry, can someone make a follow-up PR?

andrewthad:
Good catch. Also requires a major version bump since it breaks the code of anyone who expects `coerce` to work on `PrimArray`.

PR #351, opened 2022-08-02
Checked flag

andrewthad:
Add a `checked` flag to add bounds checking to element access operations. Still working on this. It does add some amount of conceptual overhead to the code (no performance overhead though).


konsumlamm:
I know this is still WIP, but I don't think duplicating the code is a good idea maintenance-wise. Almost every change would have to be made in two places then. One alternative I can think of is making a `CHECK_BOUNDS` (CPP) macro (as was done in `vector`), that is a no-op without the `checked` flag and adds bound checks with the `checked` flag. I don't see any immediate downsides of that approach (if we wanna go with a global flag), but maybe I'm missing something.

andrewthad:
Yeah, the maintenance burden is a concern. CPP is probably easier to maintain.

Bodigrim:
What's the benefit of this over `cabal test --ghc-options='-fcheck-prim-bounds'`?

andrewthad:
The GHC flag that you mention doesn’t cause any out-of-line primops to be checked. So, copyByteArray can still corrupt memory. There’s a debug version of the GHC runtime that adds checks on some (not all) of the out-of-line primops.

andrewthad:
Additionally, the GHC option causes an error that includes no information about the index or the length of the array. The approach in this PR provides that information in the error message.

andrewthad:
Not planning on picking this back up any time soon. Closing.

PR #352, opened 2022-08-03
Test GHC 9.2 in CI

konsumlamm:
Explicitly test 9.2 and update actions & `tested-with`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
       fail-fast: true
       matrix:
         os: [ubuntu-latest]
-        ghc: ['8.0', '8.2', '8.4', '8.6', '8.8', '8.10', '9.0']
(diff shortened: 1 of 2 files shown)


andrewthad:
The cabal file mentions 9.2.2, but that release is dated. GHC 9.2.4 just came out this last week, and GHC 9.2.3 has been out a while. But, it looks like 9.2.2 is what Circle CI uses. Oh well. I'll merge once the checks finish.

konsumlamm:
> The cabal file mentions 9.2.2, but that release is dated. GHC 9.2.4 just came out this last week, and GHC 9.2.3 has been out a while. But, it looks like 9.2.2 is what Circle CI uses. Oh well. I'll merge once the checks finish.

Yeah, the reason i wrote 9.2.2 is that this is the version that's actually tested.

PR #354, opened 2022-08-10
Re-export ByteArray from base-4.17

Bodigrim:


Bodigrim:
@andrewthad what do you think about this?

andrewthad:
I’ve not reviewed it, but yes, I agree with the idea of it. Since it’s passing CI, it’s almost certainly correct though. I’ll review this tomorrow.

andrewthad:
Merged. Also, I've added a changelog entry about this in 5d233dc9b1056fa5aacabfb70b9d5d961d85dc76.

PR #358, opened 2022-09-18
Fix size computation in `copyByteArrayToPtr` & `copyMutableByteArrayToPtr`

### LINKED ISSUE #345: 'copyByteArrayToPtr' uses number of bytes instead of number of elements
https://github.com/haskell/primitive/blob/b0c5294fe6351ea9c649ed2522a61362b31fa79a/Data/Primitive/ByteArray.hs#L455-L466
According to the documentation, 'sz' should be the "number of elements", while 'copyByteArrayToAddr#' wants  "number of bytes". Line 464 should be something like `... ((unI# sz) *# siz#) ...`. This same issue applies to 'copyMutableByteArrayToPtr'.

konsumlamm:
Fixes #345.


andrewthad:
Gosh, I can’t believe that’s been lurking in the code all this time. Thanks!

PR #363, opened 2023-01-13
Add keepAlive

### LINKED ISSUE #362: Add keepAlive in addition to touch?
The `keepAlive#` primop has been there for some time. Maybe it makes sense to have a `PrimMonad`-polymorphic `keepAlive`, defined similarly to `touch`. And it can work for all GHC major versions, when `keepAlive#` is not present, a stub implementation using `touch` is possible.

I can open a PR for this if `primitive` maintainers consider this a worthy idea.

TerrorJack:
Closes #362.

### DIFF
--- a/Control/Monad/Primitive.hs
+++ b/Control/Monad/Primitive.hs
@@ -22,10 +22,13 @@ module Control.Monad.Primitive (
   liftPrim, primToPrim, primToIO, primToST, ioToPrim, stToPrim,
   unsafePrimToPrim, unsafePrimToIO, unsafePrimToST, unsafeIOToPrim,
   unsafeSTToPrim, unsafeInlinePrim, unsafeInlineIO, unsafeInlineST,
-  touch, evalPrim, unsafeInterleave, unsafeDupableInterleave, noDuplicate
+  touch, keepAlive, evalPrim, unsafeInterleave, unsafeDupableInterleave, noDuplicate
 ) where
 
 import GHC.Exts   ( State#, RealWorld, noDuplicate#, touch#
+#if defined(HAVE_KEEPALIVE)
+                  , keepAlive#
+#endif
                   , unsafeCoerce#, realWorld#, seq# )
 import GHC.IO     ( IO(..) )
 import GHC.ST     ( ST(..) )
@@ -334,6 +337,14 @@ touch :: PrimMonad m => a -> m ()
 {-# INLINE touch #-}
 touch x = unsafePrimToPrim
         $ (primitive (\s -> case touch# x s of { s' -> (# s', () #) }) :: IO ())
+
+keepAlive :: PrimBase m => a -> (a -> m r) -> m r
+{-# INLINE keepAlive #-}
+#if defined(HAVE_KEEPALIVE)
+keepAlive x k = unsafeIOToPrim $ primitive $ \s0 -> keepAlive# x s0 $ internal $ unsafePrimToIO $ k x
+#else
+keepAlive x k = k x <* touch x
+#endif
 
 -- | Create an action to force a value; generalizes 'Control.Exception.evaluate'
 --
--- a/Control/Monad/Primitive.hs
+++ b/Control/Monad/Primitive.hs
@@ -339,10 +339,11 @@ touch x = unsafePrimToPrim
         $ (primitive (\s -> case touch# x s of { s' -> (# s', () #) }) :: IO ())
 
 keepAlive :: PrimBase m => a -> (a -> m r) -> m r
-{-# INLINE keepAlive #-}
 #if defined(HAVE_KEEPALIVE)
+{-# INLINE keepAlive #-}
 keepAlive x k = unsafeIOToPrim $ primitive $ \s0 -> keepAlive# x s0 $ internal $ unsafePrimToIO $ k x
 #else
+{-# NOINLINE keepAlive #-}
 keepAlive x k = k x <* touch x
 #endif
 
--- a/changelog.md
+++ b/changelog.md
@@ -12,6 +12,9 @@
     library. This does not change the user-facing interface of
     `Data.Primitive.ByteArray`.
 
+  * Add `keepAlive` that wraps `keepAlive#` for GHC 9.2 and newer. It
+    falls back to using `touch` for older GHCs.
+
 ## Changes in version 0.7.4.0
 
   * Add Lift instances (#332)
--- a/primitive.cabal
+++ b/primitive.cabal
@@ -56,6 +56,9 @@ Library
                , transformers >= 0.5 && < 0.7
                , template-haskell >= 2.11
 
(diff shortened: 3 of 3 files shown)


andrewthad:
I'll leave this open for another day or two in case anyone else with more knowledge about this primop has any feedback, and I'll merge if I don't hear anything.

mitchellwrosen:
Hey! Just curious, why does this combinator have the type

```haskell
keepAlive :: PrimBase m => a -> (a -> m r) -> m r
```

and not

```haskell
keepAlive :: PrimBase m => a -> m r -> m r
```

treeowl:
@mitchellwrosen , that's an extremely good question. It looks like a mistake..... @andrewthad ?

chessai:
It is analogous to the "with" family of combinators, e.g. withFile. keepAlive x k ensures x is kept alive as long as it is referenced by the body of k. To my knowledge, the signature suggested would not provide better guarantees than the flawed `touch`.

treeowl:
Okay, maybe there's a real idea here, but I'm not sure it's expressed well enough. The only reasons I can think of to keep a lifted value alive are to keep a weak reference active or (relatedly) to delay calling an attached finalizer. Getting that right may involve some `NOINLINE`s and/or `lazy` innovations to make sure the thing kept alive is the thing the weak pointer attaches to. Maybe this is to help with that? I don't think it goes all the way though.

chessai:
> It is analogous to the "with" family of combinators, e.g. withFile. keepAlive x k ensures x is kept alive as long as it is referenced by the body of k. To my knowledge, the signature suggested would not provide better guarantees than the flawed `touch`.

For a more thorough explanation, see https://www.haskell.org/ghc/blog/20210607-the-keepAlive-story.html

treeowl:
@chessai I think we all know what the primop is for....

chessai:
> @chessai I think we all know what the primop is for....

Well, the link goes into a detailed breakdown of why the type is the way it is. So I thought it might be helpful. Maybe I'm missing something though.

treeowl:
@chessai , I guess you're talking about the situation where the value isn't supposed to be used after its finalizer runs, like a file handle? I don't know if that's the only use case for the primop.

treeowl:
@chessai , the type of the function is not analogous to the type of the primop, which is unusual. I can imagine a reason for such a function, but as I said I don't think this realizes that exactly.

treeowl:
I'd expect the fancy version to look more like

```haskell
keepAlive x k = unsafeIOToPrim $ primitive $ \s0 -> keepAlive# x s0 $ internal $ unsafePrimToIO $ k (lazy x)
```

The `lazy` would help prevent the function argument from being unboxed/reboxed. But I'm not sure this approach will substitute for extreme care using something more like the raw primop.....

mitchellwrosen:
@chessai Thanks for the information. I (tried to) read the article you linked and I admit I didn't get exact clarity on your response:

> keepAlive x k ensures x is kept alive as long as it is referenced by the body of k

I believe the documentation of [`withForeignPtr`](https://hackage.haskell.org/package/base-4.18.0.0/docs/Foreign-ForeignPtr.html#v:withForeignPtr), which essentially just  wraps `keepAlive#`, suggests the opposite:

> The foreign object is kept alive at least during the whole action, even if it is not used directly inside.

🤔

andrewthad:
I agree that the type of `keepAlive` is essentially wrong. I should have thought about this more before merging. The type should be

    keepAlive :: PrimBase m => a -> m r -> m r

because it most closely matches the type of the primop

    keepAlive# :: v -> State# RealWorld -> (State# RealWorld -> p) -> p

but with `p` specialized to `(# State# RealWorld, q #)`. [GHC Issue 21868](https://gitlab.haskell.org/ghc/ghc/-/issues/21868) explains this weird return type, but it's irrelevant to the discussion here. Unfortunately, this has already been released, so changing it is difficult. We could introduce another function with the correct signature, soft deprecate `keepAlive`, and then either remove it or just change it in the next release of `primitive`.

@treeowl describes attaching finalizers to lifted things as "weird". I would go one step further and call it "dangerous". But it's something that GHC explicitly allows, so I think it's reasonable to support it.

I think that a useful addition would be

    keepMutableByteArrayAlive :: PrimBase m => MutableByteArray s -> m r -> m r

Which would call `keepAlive#` on the inner `MutableByteArray#`, not on the wrapper. And it would be good to have a note about why such a specialization exists.

treeowl:
@andrewthad I think it would probably be best to have a discussion of such things as part of a broader discussion of the fate of `primitive-unlifted`. There's nothing special about `MutableByteArray`; one might wish to keep anything `UnliftedType` alive.

andrewthad:
Since you raised the issue, here are my thoughts. If there's a reasonable way to do it, I think it would be nice to fold pieces of `primitive-unlifted` into `primitive`. For example, the `PrimUnlifted` class is clearly sound and works all the way back to antiquity. So I think it would be appropriate to just move it over to `primitive`. And that gives enough to build `keepAliveUnlifted` or whatever we might call it. All the stuff that has a different code path for the pre-ghc-9.4 world I'm less inclined to move into `primitive`. I've been burned a lot by `unsafeCoerce#`, sometimes by my own mistakes, sometimes by GHC bugs (since `unsafeCoerce#` doesn't get tested a whole lot by applications), and sometimes by things that I just never figured out at all.

treeowl:
`PrimUnlifted` actually isn't quite safe.

```haskell
fromUnlifted# x = fromShortByteStringUnsafe (SBS x)
```

bothers me. I'd be much happier if the `Unlifted` version of `ShortText` were a newtype around `ByteArray#` that allows for invariant checking.

andrewthad:
Yes, the instance for `ShortText` is not quite right. Used in some of my other projects, I have [Data.Text.Short.Unlifted](https://github.com/andrewthad/unlifted/blob/20b7471ac799d799ca0de192fd41d8f74f5c599d/src/Data/Text/Short/Unlifted.hs), which I [would like to move into the text-short library](https://github.com/haskell-hvr/text-short/issues/38).

chessai:
Ah, you're right. I didn't look closely at the primop before commenting, instead working off memory. This is really unfortunate. Could we do an impact analysis to determine how much it's actually used? It's so new, I doubt many are even using it. There are 4 of some of the most active users of primitive in this thread and it just got noticed. Fixing it would have to be a major version bump, unless we soft deprecated and waited, like Andrew said. Let me know if an impact assessment would be useful, and I can do it.

chessai:
> I think we should fix it and do a major version bump.

This is my preference as well

treeowl:
@andrewthad Yeah, maybe we should keep `primitive-unlifted` as as separate package. The next release of that should drop the pre-9.4 stuff, and we should put some good links in the `primitive` documentation to help people find the relevant `primitive-unlifted` stuff.

PR #372, opened 2023-03-08
Add getSizeofSmallMutableArray

treeowl:


chessai (approved):

treeowl:
@chessai I added some documentation since your review. Would you mind giving it a once-over?

konsumlamm:
So `sizeofSmallMutableArray` sometimes returns the wrong result? We should think about deprecating it then, just like the underlying primitive.

chessai (approved):

chessai:
Agree that it should be a separate issue. For now this LGTM.

treeowl:
I've opened #373 to discuss that.

treeowl:
Thanks for the review, @chessai .

PR #375, opened 2023-03-21
fix type signature of keepAlive

chessai:
Addresses the issue discussed in the conversation starting with [this comment](https://github.com/haskell/primitive/pull/363#issuecomment-1476951030).

cc @andrewthad @treeowl @mitchellwrosen for review

cc @bgamari for thoughts as well

### DIFF
--- a/Control/Monad/Primitive.hs
+++ b/Control/Monad/Primitive.hs
@@ -340,13 +340,14 @@ touch :: PrimMonad m => a -> m ()
 touch x = unsafePrimToPrim
         $ (primitive (\s -> case touch# x s of { s' -> (# s', () #) }) :: IO ())
 
-keepAlive :: PrimBase m => a -> (a -> m r) -> m r
+keepAlive :: PrimBase m => a -> m r -> m r
 #if defined(HAVE_KEEPALIVE)
 {-# INLINE keepAlive #-}
-keepAlive x k = unsafeIOToPrim $ primitive $ \s0 -> keepAlive# x s0 $ internal $ unsafePrimToIO $ k x
+keepAlive x k =
+  unsafeIOToPrim $ primitive $ \s0 -> keepAlive# x s0 $ internal $ unsafePrimToIO k
 #else
 {-# NOINLINE keepAlive #-}
-keepAlive x k = k x <* touch x
+keepAlive x k = k <* touch x
 #endif
 
 -- | Create an action to force a value; generalizes 'Control.Exception.evaluate'
--- a/changelog.md
+++ b/changelog.md
@@ -1,4 +1,6 @@
-## Changes in version ?????
+## Changes in version 0.9.0.0
+  
+  * Fix signature of `keepAlive`.
 
   * Add `getSizeofSmallMutableArray` that wraps `getSizeofSmallMutableArray#`
     from `GHC.Exts`.
--- a/primitive.cabal
+++ b/primitive.cabal
@@ -1,6 +1,6 @@
 Cabal-Version:  2.0
 Name:           primitive
-Version:        0.8.0.0
+Version:        0.9.0.0
 License:        BSD3
 License-File:   LICENSE
 


chessai:
This function needs documentation. Does anyone have any wording in mind?

mitchellwrosen:
Looks good, thanks.

But while we're making one breaking change, I think I'll propose another, though they are arguably orthogonal.

I should think that any user with a compiler older than GHC 9.0, for which this primitive did not exist, would not find a `keepAlive` symbol in this API at all. But instead, we export something with the same type signature, but subtly different and sometimes-unsafe semantics, which motivated the development of `keepAlive#` in the first place! 😬

treeowl:
@mitchellwrosen, I believe the `NOINLINE` makes that safe, because it prevents the compiler from using the fact that the action doesn't terminate (normally). Is there some additional risk?

mitchellwrosen:
Oh, yes, thank you that makes sense. I'm not totally confident a successfully smart computer couldn't conclude that `touch` is sometimes unreachable regardless, but if this tricks GHC's simplifier, that's good enough.

treeowl:
@mitchellwrosen It only has to trick sufficiently *old* versions of GHC, so we should be okay.

andrewthad:
I think this looks good. I'll merge this shortly unless there are any concerns about it. I've got something in mind for documentation, but it's going to include a lengthy description of lifted types, unlifted types, and types with identity, so I'll do that in its own PR and we can discuss there.

chessai:
Last thing I would add to this PR is that we should look into other outstanding PRs/TODOs if we're going to cut a major version release.

treeowl:
Unless you consider the primop type too conservative. Maybe it is?

treeowl:
It actually almost certainly is. @chessai, do you think you could open a GHC proposal to make `keepAlive#` work with arbitrary `State#` types? I don't imagine there's any reason to imagine such an operation wouldn't be useful and sometimes necessary in `ST`+FFI.

andrewthad:
@chessai I'm not sure why you think the new type is too liberal. What type do you think would be right?

treeowl:
@andrewthad That's what I thought for a few minutes, because it unsafely coerces the state token to work in contexts where the primop doesn't. But now I'm wondering if the primop is overly restrictive....

treeowl:
Really, there should be one decision for both the primop and this function.

chessai:
> @chessai I'm not sure why you think the new type is too liberal. What type do you think would be right?

I think @treeowl said this, unless I'm confused

andrewthad:
Sorry, I meant to address that last message to @treeowl

I now understand that @treeowl is suggesting

    keepAlive :: (PrimBase m, PrimState m ~ RealWorld) => a -> m r -> m r

to match what GHC does. However, `primitive` already disagrees with `ghc-prim` in its treatment of `touch`. GHC wants a `RealWorld` token, and `primitive` wants it to work in `ST`, not just `IO`.

Here, I think that it's GHC that should change, not `primitive`. There are non-IO situations where it makes sense to use `byteArrayContents` or `mutableByteArrayContents` to get an pointer into a byte array, and we need to keep the byte array live until the end. It's the ST+FFI situations (which sadly GHC requires unsafeCoerce, another thing that I think could stand to be improved) where I see this the most.

lehins:
@treeowl and @chessai It is totally ok for `keepAlive` to work in `ST`, it does not have to have `RealWord` state token. So the signature and implementation are absolutely fine IMHO.

In fact I've asked @bgamari to do change the signature for `keepAlive#`  when he was working on the implementation: https://gitlab.haskell.org/ghc/ghc/-/merge_requests/3131#note_274330 His response was:
>  I have no objection to generalizing over the state token type.

I guess he just probably forgot to do that.

treeowl:
Okay, please consider my concern entirely retracted. We should be sure to remove the in unsafe correction once it becomes unnecessary.

simonpj:
I have not been following the details here, but I'd love to get a thumbs-up from @bgamari .  He has spend ages getting `keepAlive#` right and knows where all the bodies are buried.

andrewthad:
I've opened [GHC Issue 23163](https://gitlab.haskell.org/ghc/ghc/-/issues/23163) to discuss there.

PR #380, opened 2023-03-23
Use mutableByteArrayContents# when available

andrewthad:


treeowl (approved):
And another one down,
And another one down;
Another one bites the dust!

andrewthad:
<insert Freddie Mercury gif here>

PR #183, opened 2018-07-05
Add complex prim instance

chessai:

PR #381, opened 2023-03-23
Add Prim instances for Data.Complex

andrewthad:
Add tests for the Complex instance. Remove usage of -XInstanceSigs because of older GHCs.

This is just https://github.com/haskell/primitive/pull/183 but squashed and rebased. I'm going to wait to merge until the `sizeOfType#` branch is merged since it will change how the `Prim` instance for `Complex` is defined.

### DIFF
--- a/Data/Primitive/Types.hs
+++ b/Data/Primitive/Types.hs
@@ -134,7 +134,7 @@ sizeOf x = I# (sizeOf# x)
 alignment :: Prim a => a -> Int
 alignment x = I# (alignment# x)
 
--- | @since 0.6.5.0
+-- | @since 0.9.0.0
 instance Prim a => Prim (Complex a) where
   sizeOf# _ = 2# *# sizeOf# (undefined :: a)
   alignment# _ = alignment# (undefined :: a)
--- a/Data/Primitive/Types.hs
+++ b/Data/Primitive/Types.hs
@@ -38,6 +38,7 @@ import Data.Primitive.Internal.Operations
 import Foreign.Ptr (IntPtr, intPtrToPtr, ptrToIntPtr, WordPtr, wordPtrToPtr, ptrToWordPtr)
 import Foreign.C.Types
 import System.Posix.Types
+import Data.Complex
 
 import GHC.Word (Word8(..), Word16(..), Word32(..), Word64(..))
 import GHC.Int (Int8(..), Int16(..), Int32(..), Int64(..))
@@ -53,7 +54,6 @@ import qualified Foreign.Storable as FS
 
 import GHC.IO (IO(..))
 import qualified GHC.Exts
-
 
 import Control.Applicative (Const(..))
 import Data.Functor.Identity (Identity(..))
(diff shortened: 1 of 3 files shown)


Data/Primitive/Types.hs
```diff
@@ -134,6 +134,47 @@ sizeOf x = I# (sizeOf# x)
 alignment :: Prim a => a -> Int
 alignment x = I# (alignment# x)
 
+-- | @since 0.6.5.0
```
konsumlamm, line 137:
```suggestion
-- | @since 0.9.0.0
```

Data/Primitive/Types.hs
```diff
@@ -134,6 +134,47 @@ sizeOf x = I# (sizeOf# x)
...
+    in x :+ y
+  readByteArray# arr# i# =
+    \s0 -> case readByteArray# arr# (2# *# i#) s0 of
+       (# s1#, x #) -> case readByteArray# arr# (2# *# i# +# 1#) s1# of
+          (# s2#, y #) -> (# s2#, x :+ y #)
+  writeByteArray# arr# i# (a :+ b) =
+    \s0 -> case writeByteArray# arr# (2# *# i#) a s0 of
+       s1 -> case writeByteArray# arr# (2# *# i# +# 1#) b s1 of
+         s2 -> s2
+  setByteArray# = defaultSetByteArray#
```
konsumlamm, line 153: Is there a reason this isn't an actual default definition? Same for `setOffAddr#`.

test/main.hs
```diff
@@ -152,6 +153,9 @@ main = do
     , testGroup "DefaultSetMethod"
       [ lawsToTest (primLaws (Proxy :: Proxy DefaultSetMethod))
       ]
+    , testGroup "Complex"
+      [ lawsToTest (primLaws (Proxy :: Proxy (Complex Double)))
+      ]
```
konsumlamm, line 158: Shouldn't this go under `testGroup "Prim"`?

test/main.hs
```diff
@@ -179,15 +183,29 @@ main = do
...
+      , lawsToTest (primLaws (Proxy :: Proxy (Down Int16)))
+      , lawsToTest (primLaws (Proxy :: Proxy (Identity Int16)))
+      , lawsToTest (primLaws (Proxy :: Proxy (Monoid.Dual Int16)))
+      , lawsToTest (primLaws (Proxy :: Proxy (Monoid.Sum Int16)))
+      , lawsToTest (primLaws (Proxy :: Proxy (Monoid.Product Int16)))
+      , lawsToTest (primLaws (Proxy :: Proxy (Semigroup.First Int16)))
+      , lawsToTest (primLaws (Proxy :: Proxy (Semigroup.Last Int16)))
+      , lawsToTest (primLaws (Proxy :: Proxy (Semigroup.Min Int16)))
+      , lawsToTest (primLaws (Proxy :: Proxy (Semigroup.Max Int16)))
+      ]
```
konsumlamm, line 197: This looks like a leftover from #183, those instance are already tested above.

test/main.hs
```diff
@@ -179,15 +183,29 @@ main = do
...
 
 deriving instance Arbitrary a => Arbitrary (Down a)
 -- Const, Dual, Sum, Product: all have Arbitrary instances defined
 -- in QuickCheck itself
+#if MIN_VERSION_base(4,9,0)
 deriving instance Arbitrary a => Arbitrary (Semigroup.First a)
 deriving instance Arbitrary a => Arbitrary (Semigroup.Last a)
 deriving instance Arbitrary a => Arbitrary (Semigroup.Min a)
 deriving instance Arbitrary a => Arbitrary (Semigroup.Max a)
+#endif
```
konsumlamm, line 208: Same for this, our minimum `base` version is 4.9.

konsumlamm (commented):
I'm in favor, `Complex` is a pretty basic type and I expect many want to use it in `PrimArray`s (especially for numerical computations).

konsumlamm:
If we merge this, we'll have to think about wether we also want instances for tuples. The instances are similarly easy to implement. Compare this with [Data.Vector.Unboxed](https://hackage.haskell.org/package/vector/docs/Data-Vector-Unboxed.html), where arrays of tuples (and `Complex`) are implemented as tuples of arrays.

lehins:
@konsumlamm Tuples have different types, while Complex has a single type.

Vector's `Unbox`ing is implemented with Structure of Arrays approach, while `Prim` is Array of structures. 

There is a whole lot of worms in adding instances of `Prim` for tuples.

For example today we can add instances like this:

````
instance Prim a => Prim (a, a) where
...

instance Prim a => Prim (a, a, a) where
...
```

But not like this:
```
instance (Prim a, Prim b) => Prim (a, b) where
...
```
because `Prim` does not have the ability to index `ByteArray#`/`MutableByteArray#` at a byte offset. There are fairly recent primops that allow this, but `Prim` class hasn't been modernized to use them.

There are also questions of alignment. For example what is the alignment for `Prim (Word64, Word8`)?

andrewthad:
Thanks for the feedback on the PR. I'll address those soon. Concerning tuples, I agree with @lehins. The only way to define a reasonable instance for tuples would be

    instance (Prim a, a ~ b) => Prim (a,b) where

If we let the elements be of different types, then we have to worry about alignment and padding, and it's not possible to pack things well with Haskell's typeclasses. You would need something stronger like metaprogramming.

konsumlamm:
@lehins those are very good points, in that case I don't think we should add instances for tuples, at least for now.

Data/Primitive/Types.hs
andrewthad: That's a good point. I'm going to make this the default in a separate PR and then rebase this on top.

konsumlamm:
Another point in favor of this is that there's also `instance Storable a => Storable (Complex a)`.

konsumlamm:
BTW the title is misleading, the PR only adds `Prim` instances for `Data.Complex`.

andrewthad:
I've updated the title on the PR. I'll make sure to correct the commit message too.

Data/Primitive/Types.hs
andrewthad: This has been resolved at https://github.com/haskell/primitive/pull/385.

andrewthad:
I've taken care of the outstanding issue and put up a rebased version of this at https://github.com/haskell/primitive/pull/385.

PR #390, opened 2023-08-12
Clarify that newly allocated `ByteArray`s/`PrimArray`s are uninitialized

konsumlamm:
I just had a bug, because I assumed `ByteArray`s/`PrimArray`s were initialized with zeros. Unfortunately, the documentation didn't mention that they're uninitialized (except for `newPrimArray`).


andrewthad:
Thanks! Definitely worth calling this out explicitly in the docs. The worst part is that in something super short running, it ends up filled by zeroes (because the memory pages GHC's runtime requests from the kernel with `mmap` is zero filled), but then after the first GC, once pages start getting reused, then you get burned.

PR #398, opened 2023-08-29
Make it build with ghc-9.8

erikd:
Only makes the library itself build. The test suite has broken dependencies, some of which in turn have broken dependencies.


andrewthad:
I've opened https://github.com/haskell/primitive/issues/399 to track the inability to build the test suite.

PR #407, opened 2024-01-04
Use `copyMutableByteArray#` in moveByteArray

clyring:
The primop in question appears to have always been safe for use with overlapping ranges since its introduction. See also: https://gitlab.haskell.org/ghc/ghc/-/commit/b00b36196a88ad6b9054244caaec926f6f9db2cf

The cbits function "hsprimitive_memmove" is now unused, but is not removed because it was arguably publicly exposed due to the "Install-Includes: primitive-memops.h".


andrewthad:
Looks good. Merging.

Issue #265, opened 2020-03-09
memset is terrible and not performant likely (for nonzeroing path)


cartazio:
it falls back to a scalar c loop when not zeroing. which seems disappointing. Also not sure if it correctly handles `-0` in the floating point code

cartazio:
this is gonna be some fun hackery to do a sane portable performant thingy :)

cartazio:
@bgamari under what circumstances does the GHC code gen do something smart with memset/memmove? As best i can recall, the only instances are internally generated and via the llvm codegen?

andrewthad:
GHC has a primop for `memset` (`setByteArray#`) that is not currently used by `primitive`. There is a [PR that switches to using this](https://github.com/haskell/primitive/pull/222). In GHC 8.10, GHC is going to become smarter about unrolling this when the size is known and small.

andrewthad:
Original issue on GHC is [here](https://gitlab.haskell.org/ghc/ghc/issues/16052) and Artem's work that resolved the issue is [here](https://gitlab.haskell.org/ghc/ghc/commit/7f33979d0160c803a215dd166d90d4e57e1f6d42).

cartazio:
I’ll look at how ghc does that primop and figure out what the different choices are.  Thanks for helping out :)

cartazio:
ok, so 
https://gitlab.haskell.org/ghc/ghc/blob/4898df1cc25132dc9e2599d4fa4e1bbc9423cda5/compiler/GHC/StgToCmm/Prim.hs#L2480-2496
```haskell
-- ----------------------------------------------------------------------------
-- Setting byte arrays

-- | Takes a 'MutableByteArray#', an offset into the array, a length,
-- and a byte, and sets each of the selected bytes in the array to the
-- character.
doSetByteArrayOp :: CmmExpr -> CmmExpr -> CmmExpr -> CmmExpr
                 -> FCode ()
doSetByteArrayOp ba off len c = do
    dflags <- getDynFlags

    let byteArrayAlignment = wordAlignment dflags -- known since BA is allocated on heap
        offsetAlignment = cmmExprAlignment off
        align = min byteArrayAlignment offsetAlignment

    p <- assignTempE $ cmmOffsetExpr dflags (cmmOffsetB dflags ba (arrWordsHdrSize dflags)) off
    emitMemsetCall p c len align

```

and `emitMemsetCall` is defined as 
```haskell
-- | Emit a call to @memset@.  The second argument must fit inside an
[...]

PR #433, opened 2025-12-17
Fix `setPrimArray` for `Float` & `Double`

konsumlamm:
Fix the bug mentioned in #265 by generating different functions for `Float` and `Double` that check `x == 0.0 && !signbit(x)` to ensure that we only use `memset` for `+0.0` (and not `-0.0`).

I also removed the special case for `sizeof(Hs ## TYPE) == sizeof(int)*2` since it didn't yield better performance in my benchmark. I also don't see why one would expect performance improvements from that in the first place, since a single access should be faster than two.

### DIFF
--- a/cbits/primitive-memops.c
+++ b/cbits/primitive-memops.c
@@ -1,3 +1,4 @@
+#include <math.h>
 #include <string.h>
 #include "primitive-memops.h"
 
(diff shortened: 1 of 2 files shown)


andrewthad (approved):

andrewthad:
This looks correct to me. And also, it looks like the tests suite would catch a mistake in the implementation of Prim for Float or Double. Are you alright with me merging this?

treeowl:
> Yes, I added `primLaw` tests for `Float` and `Double` since they were missing. I also added separate tests for the `-0.0` bug, since as far as I can tell the `Arbitrary` instances don't generate `-0.0`.

Should that be considered a bug in `QuickCheck`? Has it been reported?

PR #445, opened 2026-09-07
Fix signed fill arguments at the C FFI boundary

felixonmars:
The setInt8/16/32Array# and setInt8/16/32OffAddr# imports pass signed values to C helpers that expect unsigned arguments. On RISC-V, the calling convention requires different extension for signed and unsigned 16-bit values. Filling four Int16 elements with -2 can therefore produce [-2,-1,-1,-1].

For GHC 9.2 and newer, convert the narrow signed values to their unsigned equivalents and call the existing unsigned imports through inline wrappers. This preserves the element bits while matching the C argument types.

This fixes 13 failures in the existing test suite on riscv64: Int16 filling, replicatePrimArray, and ByteArray Set Range for Int16 and its newtype wrappers.

### DIFF
--- a/Data/Primitive/Internal/Operations.hs
+++ b/Data/Primitive/Internal/Operations.hs
@@ -70,12 +70,20 @@ foreign import ccall unsafe "primitive-memops.h hsprimitive_memset_Word"
   setWordArray# :: MutableByteArray# s -> CPtrdiff -> CSize -> Word# -> IO ()
 
 #if __GLASGOW_HASKELL__ >= 902
-foreign import ccall unsafe "primitive-memops.h hsprimitive_memset_Word8"
-  setInt8Array# :: MutableByteArray# s -> CPtrdiff -> CSize -> Int8# -> IO ()
-foreign import ccall unsafe "primitive-memops.h hsprimitive_memset_Word16"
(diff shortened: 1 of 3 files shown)


konsumlamm:
What about GHC < 9.2? Is the bug not present there?

felixonmars:
> What about GHC < 9.2? Is the bug not present there?

I don't really have the environment to test because GHC < 9.2 isn't very usable on riscv64. I just added a commit for that branch as well so you can decide whether it's worth it :)

andrewthad:
Nice fix. The previous approach was already on shaky footing to begin with since it used the FFI to wrap a C function with the wrong type (implicitly casting `Int16#` to `Word16#` for example). The popular ABIs for x86 and ARM include calling conventions that allow you to put garbage in the upper bits of sub-machine-sized word types when you pass them to functions, so this always just kind of worked out before.

The only thing that I'm not sure of is whether or not this library should still be supporting GHCs earlier than 9.2. It would be nice to just drop all of the shims. I've opened https://github.com/haskell/primitive/issues/446 for a discussion on this. Before merging this, I'd like to see if it's possible to quickly gather consensus about dropping support for old GHCs (in the linked issue). If it seems like it will take more time to figure that out, I'll merge this as is. But if there is strong consensus that old GHCs are not worth the effort, we should drop the compatibility paths from this before merging.

